PR #16, opened 2016-09-15
[#15] problems when installing the extension in chrome

### LINKED ISSUE #15: problems when installing the extension in chrome
Sadly the zip file that is created in the dist directory does not include the images and chrome complains when installing it:

bildschirmfoto 2016-09-15 um 21 35 16

bitboxer:
fixes #15

### DIFF
--- a/gulpfile.babel.js
+++ b/gulpfile.babel.js
@@ -156,7 +156,7 @@ gulp.task('build:webext', [
   'build:webext:icons',
   'build:webext:manifest'
 ], () => {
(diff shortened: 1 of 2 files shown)


bitboxer:
I think the zipfile is now working correctly. @pmeinhardt please check.

bitboxer:
(After this is merged, the version number needs to increase to 1.0.1)

pmeinhardt:
Awww, yes. Thanks a lot for spotting this and creating a PR!

Issue #12, opened 2016-06-08
Bugs/Chores are not recognized as such in JIRA/Atlassian


andreasknoepfle:
Somehow in JIRA every generated branch seems to start with `feature/ ` no matter if its a feature, a bug or a chore ticket. 
(can be reproduced in Chrome and Firefox)

bitboxer:
This still happens in 1.0.1

pmeinhardt:
Done, via #32

pmeinhardt:
Nice work @weiland

PR #32, opened 2016-09-26
[#12] Bugs/Chores are not recognized as such in JIRA/Atlassian

weiland:


bitboxer:
Is this PR working for ticket pages and the dashboard? I thought it would need 2 different approaches there.

weiland:
In the end I could simplify the code since there were already two selectors.

bitboxer:
Okay. Perfect. @pmeinhardt can you check this and merge if okay?

spec/adapters/jira-spec.js
```diff
@@ -10,6 +10,11 @@ const TICKETPAGE = `
       <a class="issue-link" data-issue-key="UXPL-39" id="key-val">UXPL-39</a>
       <h1 id="summary-val">A Random JIRA Issue</h1>
     </div>
+    <div class="issue-body-content">
+      <span id="type-val" class="value editable-field inactive">
+        <img alt="" src="/" title="Bug - A problem"> Bug
+      </span>
+    </div>
```
pmeinhardt: How about adding a separate test case and HTML snippet (`BUGPAGE`) for this, so the tests also make sure the JIRA adapter distinguishes `bug` from `feature`? Like in the GitHub adapter for instance:

- https://github.com/bitcrowd/tickety-tick/blob/master/spec/adapters/github-spec.js#L14-L24
- https://github.com/bitcrowd/tickety-tick/blob/master/spec/adapters/github-spec.js#L69-L75

spec/adapters/jira-spec.js
```diff
@@ -26,7 +31,7 @@ describe('jira adapter', () => {
   });
 
   it('extracts tickets from a ticket page', () => {
-    const expected = [{ id: 'UXPL-39', title: 'A Random JIRA Issue' }];
+    const expected = [{ id: 'UXPL-39', title: 'A Random JIRA Issue', type: 'bug' }];
```
pmeinhardt: See comment on line 17.

src/common/adapters/jira.js
```diff
@@ -8,13 +15,13 @@ const adapter = {
       // TODO: check what this looks like in JIRA, update tests, find better selectors
```
pmeinhardt: I think it's safe to remove this comment now 😉

spec/adapters/jira-spec.js
```diff
@@ -2,41 +2,91 @@ import { jsdom } from 'jsdom';
...
-// const SIDEBAR = ``;
+const SIDEBAR = `<div>
+  <div class="ghx-fieldname-issuekey">
+    <a href="#">UXPL-39</a>
+  </div>
+  <div data-field-id="summary">A Random JIRA Sidebar Issue</div>
+  <div data-issue-key="UXPL-39">
+    <span class="ghx-type" title="Story"></span>
+  </div>
+</div>`;
```
pmeinhardt: Could you please make the string formatting here consistent with the rest of the HTML snippets?

I.e. everywhere else, we do

```javascript
const SNIP = `
  <div>
    …
  </div>
`;
```

so opening and closing tags are indented the same.

spec/adapters/jira-spec.js
```diff
@@ -2,41 +2,91 @@ import { jsdom } from 'jsdom';
...
+    });
+  });
+
+  it('extracts chore tickets from the sidebar', () => {
+    const expected = [{ id: 'UXPL-39', title: 'A Random JIRA Sidebar Issue', type: 'chore' }];
+    adapter.inspect(null, doc(CHORE_SIDEBAR), (err, res) => {
+      expect(err).toBe(null);
+      expect(res).toEqual(expected);
+    });
+  });
```
pmeinhardt: 👍 Thanks for adding the extra tests!

src/common/adapters/jira.js
```diff
@@ -1,5 +1,12 @@
 import { $find, $has, $text, $attr } from './helpers';
 
+const TYPES = ['bug', 'chore'];
+const normalizeType = (type) => {
+  const sanitizedType = type && type.toLowerCase();
+  if (TYPES.indexOf(sanitizedType) > -1) return sanitizedType;
```
pmeinhardt: `TYPES.includes(…)` might be nicer but with regard to browser support, your solution is likely better 😁

Issue #23, opened 2016-09-16
Add more tests for src/common/popup/utils/format.js


bitboxer:
We need more tests for the `format.js` . We have to make sure the branch names stay consistent through different versions and refactorings.

pmeinhardt:
Done, via #33. Nice work @weiland

PR #33, opened 2016-09-27
[#23] Add more tests for src/common/popup/utils/format.js

weiland:


spec/popup/format-spec.js
```diff
@@ -0,0 +1,51 @@
...
+    });
+  });
+
+  describe('normalize', () => {
+    it('formats normal string', () => {
+      const formatted = format.normalize('hello');
+      expect(formatted).toBe('hello');
+    });
+
+    it('de-capitalizes', () => {
```
pmeinhardt: It's a little more general than just `de-capitalizing`… It should really **lowercase** everything, not just the first letter.

spec/popup/format-spec.js
```diff
@@ -0,0 +1,51 @@
...
+      const formatted = format.normalize('Bitcrowd');
+      expect(formatted).toBe('bitcrowd');
+    });
+
+    it('formats a string with a space', () => {
```
pmeinhardt: Maybe more general: `it('formats strings with spaces', …)`? Instead of just `a space`.

spec/popup/format-spec.js
```diff
@@ -0,0 +1,51 @@
...
+      const formatted = format.normalize('hello bitcrowd');
+      expect(formatted).toBe('hello-bitcrowd');
+    });
```
pmeinhardt: It might be a good idea to add tests to check that special accented characters are handled properly. For instance:

- `Señor Dévèloper` should be normalized to `senor-developer`
- `äöüß` should become `aeoeuess`

pmeinhardt: Special characters often used in issue titles should be replaced. For example:

Given `[#23] Add more tests for src/common/popup/utils/format.js` as an input…

- slashes should be replaced by dashes
- as well as square brackets
- hash symbols
- and also parentheses `(` and `)`

pmeinhardt: This could be a couple of small test cases

pmeinhardt: In the end, it might be simpler to check that normalize passes the input to `speakingurl` and `slugifies` it. But it might still make sense to test for a few commonly used input characters/combinations.

spec/popup/format-spec.js
```diff
@@ -0,0 +1,51 @@
...
+  });
+});
+
```
pmeinhardt: Remove the extra new-line here

spec/popup/format-spec.js
```diff
@@ -0,0 +1,51 @@
...
+  describe('commit', () => {
+    it('formats', () => {
+      const formatted = format.commit(ticket);
+      expect(formatted).toBe(`[#${ticket.id}] ${ticket.title}`);
+    });
+  });
+
+  describe('branch', () => {
+    it('formats', () => {
```
pmeinhardt: How about `it('includes ticket type, id and title', …)`? Might be a little more descriptive?

spec/popup/format-spec.js
```diff
@@ -0,0 +1,51 @@
...
+      const formatted = format.branch(ticket);
+      expect(formatted)
+        .toBe(`${ticket.type || 'feature'}/${ticket.id}-${format.normalize(ticket.title)}`);
```
pmeinhardt: Get rid of the ` || 'feature'` here. Looks like this is copied from the implementation.

spec/popup/format-spec.js
```diff
@@ -0,0 +1,51 @@
...
+
+describe('format util', () => {
+  const ticket = {
+    id: 'BTC-042',
+    title: 'Weird Bug while eating cake',
+    type: 'bug'
+  };
+
+  describe('commit', () => {
+    it('formats', () => {
```
pmeinhardt: How about `it('includes ticket id and title', …)` to make the description more readable?

spec/popup/format-spec.js
```diff
@@ -0,0 +1,51 @@
...
+
+  describe('branch', () => {
+    it('formats', () => {
+      const formatted = format.branch(ticket);
+      expect(formatted)
+        .toBe(`${ticket.type || 'feature'}/${ticket.id}-${format.normalize(ticket.title)}`);
+    });
+
+    it('formats type to feature if not provided', () => {
+      const miniTicket = {
```
pmeinhardt: How about `typeless…` or `untyped…` or something similar for a better name?

spec/popup/format-spec.js
```diff
@@ -0,0 +1,51 @@
...
+        id: ticket.id,
+        title: ticket.title
+      };
+      const formatted = format.branch(miniTicket);
+      expect(formatted).toBe(`feature/${ticket.id}-${format.normalize(ticket.title)}`);
+    });
+  });
+
+  describe('normalize', () => {
+    it('formats normal string', () => {
```
pmeinhardt: `formats normal strings` would be grammatically correct 😁

spec/popup/format-spec.js
```diff
@@ -0,0 +1,51 @@
+import format from '../../src/common/popup/utils/format.js';
+
+describe('format util', () => {
+  const ticket = {
+    id: 'BTC-042',
+    title: 'Weird Bug while eating cake',
```
pmeinhardt: Maybe use a more realistic title? E.g. `'Add more tests for src/common/popup/utils/format.js'`

Issue #34, opened 2016-09-28
CLI command to generate commit message from branch name


PragTob:
When working with tickety-tick and I want to commit my changes I have to break out of my working context (IDE + console), go to the issue tracker, and click 2 times.

What could be a better solution (imo) would be that there was a CLI command to create a commit message from the branch name (all necessary informaion should be there).

e.g. 

```
tobi@airship ~/projects/cool $ git status
On branch feature/foo-5264-implement-cool-stuff
nothing to commit, working directory clean
tobi@airship ~/projects/cool $ tt-commit
tt-commit: command not found
tobi@airship ~/projects/cool $ # makes the commit and now I can ammend it or even better it opens my editor with the commit message prefilled
```

no idea if there should be some gem/npm package to install or just a bash snippet in the readme so that one can set up a custom alias/command.

Would improve the workflow imo :)

bitboxer:
I also thought of this, but it will be a bit harder to implement. Basically we need to store the original story name somewhere with its id to make this possible. We can't calculate that back from the branch name. Alternatively we need to ask the api of the ticket system for this information.

Sadly we can't communicate from the command line with the extension running in the browser (well, with the old firefox style plugins this was possible, but mozilla will deprecate that soonish).

So if anyone has an idea how to implement this, I am all up for this.

pmeinhardt:
One thing that comes to mind is: https://github.com/nvie/gitflow

While this creates feature/bug/chore… branches, it - however - does not integrate with any ticket tracking system (plus it's been inactive for 4 years now).

For this to work, you'd really need a per-project (git) config containing a tracker specifier, project id and personal API access token:

- This one looks like it could work nicely with Pivotal projects at least: https://github.com/nebhale/git-pivotal-tracker-integration
- Then, there's also: https://github.com/trydionel/git-pivotal
- Maybe there are more integrations for other tracking systems but I am not aware of any git CLI tool that can handle a good range of ticket/story tracking systems

pmeinhardt:
**A crazy thought**: How about building a ticket tracking extension into git? So that when you check out a repo, at the same time you check out issues/tickets… and they're updated/synchronized with your peers whenever you pull/push. I'll need to think about this a little.

andreasknoepfle:
can't we just assure, that we only generate commit/branch -names that can be inferred from each other ? like this a simple tool without any api could do the job :)

pmeinhardt:
[...]

PR #36, opened 2016-10-06
[#34] CLI command to generate commit message from branch name

weiland:
This PR was already created with this modification.


bitboxer:
Can you add an explanation for this in the readme?

src/common/popup/utils/format.js
```diff
@@ -14,8 +14,13 @@ function branch(ticket) {
   return `${ticket.type || 'feature'}/${ticket.id}-${normalize(ticket.title)}`;
 }
 
+function command(ticket) {
+  return `git checkout -b ${branch(ticket)}  && git commit --allow-empty -m"${commit(ticket)}"`;
```
pmeinhardt: There should be a space between `-m` and the message. It'll work like this, but for extra style points… 😝 (https://git-scm.com/docs/git-commit) (Sorry for nitpicking).

tessi: i gave a bad example in my proposal-issue ;) (for some reason i never write that space, to save a byte)

pmeinhardt:
Unfortunately I could not find a JavaScript package, that implements
shell-escaping. The `shellquote` function implemented here currently
does the following:

  1. Wraps argument in single-quotes (or returns `''` if none given)
  2. Escapes any single-quotes in the input string

Shell-quoting is applied to both the branch name and the commit message
generated by Tickety-Tick.

I do not at all like implementing such a crucial function here and will
do some more research into available candidate implementations & further
potential issues (e.g. Unicode).

@bitboxer, @tessi: Maybe you have some further knowledge on this matter?

I've looked at what Ruby's [Shellwords#escape] does (though not directly applicable) and a couple of other related pieces of code (e.g. [mathiasbynens/jsesc](https://github.com/mathiasbynens/jsesc)).

Bash manual on quoting: https://www.gnu.org/software/bash/manual/html_node/Single-Quotes.html

pmeinhardt:
Here's an issue that you can play around with (if you're building the extension from source).

bitboxer:
I think it's okay enough for now. Don't see any other JS library, too.

PR #45, opened 2016-10-18
[#44] Tickety is 'having' "problems" with correctly escaping 'quotes'

### LINKED ISSUE #44: Tickety is 'having' "problems" with correctly escaping 'quotes'
When using the new shell-command button, tickety does not correctly escape quotes. For example, the shell command for this issue is:

```
git checkout -b 'feature/44-tickety-is-having-problems-with-correctly-escaping-quotes' && git commit --allow-empty -m '[#44] Tickety is '\''having' "problems" with correctly escaping 'quotes''
```

Only the first `'` is escaped.

The expected (commit-part of the) command is:
```
git commit --allow-empty -m '[#44] Tickety is '\''having'\'' "problems" with correctly escaping '\''quotes'\'''
```

pmeinhardt:
Fixes #44. Thanks for reporting @tessi!

Background:

JavaScript `String.prototype.replace` can be called with a String or a RegExp as the first (match) argument. I expected the version with a String for matching to replace all occurrences, but apparently my memory tricked me.

Fix:

```js
s.replace('\'', '\'\\\'\'')
```

is now

```js
s.replace(/'/g, '\'\\\'\'')
```

Reference:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace

### DIFF
--- a/spec/popup/format-spec.js
+++ b/spec/popup/format-spec.js
@@ -58,7 +58,9 @@ describe('format util', () => {
     });
 
     it('escapes any single-quotes in the input', () => {
-      expect(format.shellquote('esc\'; echo "pwned"')).toBe('\'esc\'\\\'\'; echo "pwned"\'');
+      const input = 'you\'; echo aren\'t "pwned"';
+      const quoted = '\'you\'\\\'\'; echo aren\'\\\'\'t "pwned"\'';
+      expect(format.shellquote(input)).toBe(quoted);
     });
   });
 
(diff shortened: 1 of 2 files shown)


bitboxer:
@weiland please review and merge if you think it is okay. I will release a fixed version after that.

tessi:
LGTM (although, what do I know of JS  `¯\_(ツ)_/¯` )? But thanks to everyone for fixing the issue that fast!

PR #51, opened 2017-04-01
Elm

pmeinhardt:
Trying out Elm here since @paulodiniz got me really excited about it! 😀


paulodiniz:
This is pure awesomess ❤️

pmeinhardt:
Implementation is done. (Still want to look into testing in Elm)

A few unfiltered observations:

1. I very much like Elm's *pure*, functional concepts
  - It's something that Redux aspires to, but I was always repelled by the amount of boilerplate needed esp. for Redux actions (action type const, action creators…), connecting components…
  - In Elm all of this is expressed more concisely because it is part of the language/platform
2. typing is part of the language and not "artificially added" like PropType validations in React (which only trigger on runtime and only in development)
3. it is great to have language concepts like union types, Maybes… at hand
4. immutability out of the box!
5. Elm-JS interop may require you to jump through a couple of hoops and potentially write more code
  - You cannot easily make synchronous calls into JS and immediately use the JS return value (e.g. for using formatting/data transformation libraries or packages like octicons)
  - Instead, using ports is the way to go ("native" modules should be exceptions)
  - Thus round-tripping through an outbound, then inbound port may be necessary to update the Elm model with the result
  - This may not be as much of an "issue" when you're asynchronously fetching results from an API… anyway?
  - I avoided such roundtrips by doing all JS processing upfront, before passing values on to the Elm program
  - values that were previously dynamically re-computed are now computed once, before they reach Elm-land, which is a different approach/way of thinking
6. Comparing a few stats on the Elm and React versions:
  - *Take these numbers with a grain of salt! We need to understand the how/why… to draw conclusions. None of the versions have been optimized.*
  - `tar -cf [VERSION].tar dist/web-extension && wc -c [VERSION].tar`
    - `VERSION` = `React`: `2461184`
    - `VERSION` = `Elm`: `935936` (~38%)
  - React `popup/popup.js`: `2.1 MB`
  - Elm `popup/{popup.js,main.js}`: `456 KB + 106 KB`
7. The Elm community seems rather small
  - for instance, you cannot easily find a `speakingurl` package pendant in Elm, so you may be forced to interface with JS if you don't intend to do a complete reimplementation
  - finding developers that can read and write Elm code are likely harder to find than JS developers
  - it might also be less approachable for most server-side/backend developers that have to make client-side changes from time to time?

@paulodiniz, feel free to pitch in and add your impressions, point out things that are missing/incorrect.

pmeinhardt:
Another minor thing: The Elm `Html*` DSL feels somewhat easier to type than HTML/JSX (e.g. no need for a hard-to-reach `=` between attribute key and value pairs) and less verbose (e.g. no repetitive closing tags `</…>` needed).

pmeinhardt:
Open questions: What alternative patterns for structuring larger Elm applications are there? Best practices?

paulodiniz:
Good points on Elm:

* It's very opinionated on `how` browser applications work in general, and it gets it right. 
* The Signal mapping is a direct consequence of that. You can map different actions and events on signals and then deal with different signals on your application
* I really like the HTML DSL as also, but I sometimes get a little lost on the params. What helps me on that situation is trying to keep the html really small and concise.
* The general ideia is to having multiple components, each one with it's own main/update/etc functions. This is the idea that Elm was created for but I never created an app that needed that much components.
* It's purely functional

 Not so good points:

* Not so many libraries
* Major changes on version `0.17` which generated a lot of code on the internet that simply is not true anymore.
* It's purely functional (yes, good and bad). This is bad in the sense that sometimes the code is very different from what we're used to it so maybe onboarding is harder.


I LOVE Haskell/Elm syntax, but I think React gets most of the functional idea and also has a lot of traction for the community and big companies (way more than Elm). This implies that although Elm code can be nicer, maybe it's still too early to push it on a big project.

What do you think?

tessi:
This is not meant to be merged, right?
If so, I'd vote to close this PR (which shall be no valuation of the marvelousness (is that even a word?) of  elm)

pmeinhardt:
😭 was a lot of fun anyways 🙂 wish we could use elm in more projects…

PR #52, opened 2017-08-02
Language agnostic URLs.

marsjaninzmarsa:
Removes hardcoded language slugs (`de` and `en`) from AMO and Opera Addons URLs.


tessi:
Hey Kuba,

thanks for pointing this out and fixing it :) It seems that @bitboxer is not active on this repo at the moment, so I silently take over (this PR) and merge.

cheers,
tessi

PR #58, opened 2017-10-07
[#57] Stopped working with one of the recent gitlab updates

### LINKED ISSUE #57: Stopped working with one of the recent gitlab updates
One of the recent gitlab updates seemed to change something in their DOM bad enough to break the gitlab adapter. Currently, tickety-tick cannot find a ticket on the page.

Example ticket: https://gitlab.com/gitlab-org/gitlab-ce/issues/1

tessi:
* attempt to fix #57 (tested it manually in a recent chrome and it seems to work on gitlab issue pages again)
* needed to update jsdom to access the `dataset` api in tests. including little fixes because of jsdom updates
* implemented the one leftover pending test for gitlab

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "gulp-uglify": "^2.0.0",
     "gulp-zip": "^3.2.0",
     "jasmine": "^2.5.2",
-    "jsdom": "^9.6.0",
+    "jsdom": "^11.3.0",
     "nodemon": "^1.11.0",
     "nyc": "^10.0.0",
     "react-addons-test-utils": "^15.3.2",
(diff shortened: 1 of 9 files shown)


pmeinhardt:
@tessi, the code changes look very reasonable to me. One additional question this raises is whether we want to support different versions of trackers in parallel, e.g. going forward still support older versions of GitLab which people may be running.

On the other hand as long as there's no request for it, it's great to support just the latest version properly I think. Thanks 🤘

tessi:
With our somewhat limited resource I'm a fan of only supporting the last gitlab version (-> whatever is deployed in gitlab.com)

Issue #54, opened 2017-10-05
Extension can't be installed on new Firefox Quantum


andreasknoepfle:
As the manifest format will change for the new Firefox Quantum we need to change accordingly. Currently it is impossible to install the extension even manually in the upcoming release of Firefox:

![screen shot 2017-10-05 at 16 45 33](https://user-images.githubusercontent.com/239398/31233315-a868ff1a-a9ec-11e7-8e00-2affdbd2a4be.png)

andreasknoepfle:
The master branch is ready for Firefox Quantum now!

Issue #50, opened 2017-03-24
New Firefox release?


PragTob:
The firefox release is still [version 0.1.4](https://addons.mozilla.org/de/firefox/addon/tickety-tick/) and not the new fancy versions, which starts to get problematic as part of the branch name generation seems to have changed (the branch in newer versions seems to contain capital letters, while mine only has lower case letters) :)

Thanks!

bitboxer:
I tried several times last year to send this version to firefox, but the review process is really bad. A review said that we sent in the wrong code for a different plugin and then other reviewers irgnored the readme and tried to create the packaged version themself from the repository....I basically stopped trying. And I am a firefox user myself 😭

bitboxer:
But if you want, you can go forward and try your luck with it.

pmeinhardt:
We need to fix this! The amazing @andreasknoepfle has agreed to look into it next week 🎊

PragTob:
Also seems like firefox has changed the way they want addons to be built (again!) and so ticketytick will [stop working with Firefox 57/November this year](https://support.mozilla.org/en-US/kb/firefox-add-technology-modernizing?as=u&utm_source=inproduct)

![selection_089](https://user-images.githubusercontent.com/606517/31341595-41743740-ad0a-11e7-8ad6-267aee5624d7.png)

bitboxer:
@PragTob luckily this is already using the new web extension api.

andreasknoepfle:
https://addons.mozilla.org/de/firefox/addon/tickety-tick/
🎈 🎉  1.2.2 is here

pmeinhardt:
Whoop, whoop! 🎆

PragTob:
Thanks :tada:

bitboxer:
![](https://media.giphy.com/media/QbkL9WuorOlgI/giphy.gif)

PR #61, opened 2017-10-13
[#54] Extension can't be installed on new Firefox Quantum

pmeinhardt:
Adds an extension ID as recommended by MDN:
https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Alternative_distribution_options/Sideloading_add-ons#Preparing_your_add-on

### DIFF
--- a/src/web-extension/manifest.json
+++ b/src/web-extension/manifest.json
@@ -4,6 +4,11 @@
   "version": "#{VERSION}",
   "description": "#{DESCRIPTION}",
   "icons": {},
+  "applications": {
+    "gecko": {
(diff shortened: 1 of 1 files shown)


pmeinhardt:
FYI @andreasknoepfle I was able to install a local build of the extension in Firefox Quantum (DeveloperEdition). You have to select the built `manifest.json` (not the `zip`) for loading it as a temporary add-on.

pmeinhardt:
Maybe we can close #54 while - at the same time - taking another stab at releasing a new official Firefox version of the add-on #50.

andreasknoepfle:
Tried it. Works as a charm. Nice work :)

PR #62, opened 2017-10-13
Update dependencies

pmeinhardt:
Update most of Tickety-Tick's direct and build/test dependencies.

There's still two deprecation warnings on `yarn run checks` – potentially from `react-router`?
Other than that everything still lints, builds and works (tested in Firefox).


.eslintrc.json
```diff
@@ -1,4 +1,4 @@
 {
   "root": true,
-  "extends": "bitcrowd-react"
+  "extends": "airbnb"
```
tessi: 👍

src/common/popup/components/about.jsx
```diff
@@ -18,7 +18,7 @@ function About() {
               <img
                 src="../icons/icon-32.png"
                 className="logo-sm"
-                role="presentation"
+                alt=""
```
tessi: what was the purpose of `role` before? why do we need an empty `alt` attribute?
Shouldn't `alt`'s always have content describing the image? (I'm not sure on all these questions, its not meant as passive aggressive questions :) )

pmeinhardt: https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/alt-text.md#how-to-resolve

pmeinhardt: Since it's an icon, supposedly it's okay to set an empty `alt` attribute according to `eslint-plugin-jsx-a11y`.

pmeinhardt: You are all over the place recently @tessi 😁

tessi: :D 
if you notice that, you must be all over the place too ;)

pmeinhardt: Nah, I just get all your GitHub notifications 😛

src/common/popup/components/external-link.jsx
```diff
@@ -1,20 +1,30 @@
...
-  // eslint-disable-next-line jsx-a11y/anchor-has-content, jsx-a11y/no-static-element-interactions
-  return (<a {...props} target="_blank" rel="noopener noreferrer" onClick={handler} />);
+  /* eslint-disable jsx-a11y/anchor-has-content, jsx-a11y/no-static-element-interactions */
+  return (
+    <a
+      target="_blank"
+      rel="noopener noreferrer"
+      onClick={handler}
+      onKeyDown={handler}
+      {...props}
```
andreasknoepfle: maybe we can add `role="link"` here and remove the `no-static-element-interactions` ignore.

Also when adding `props.children` as required and explicitly setting that into the body of the `<a>` tag, we could probably remove the other eslint disable and make sure that this element always has a body of some sort.

gulpfile.babel.js
```diff
@@ -207,12 +207,12 @@ gulp.task('build:safari', [
...
   'build:safari:js',
   'build:safari:icons',
-  'build:safari:plist'
+  'build:safari:plist',
 ]);
 
 gulp.task('build', [
   'build:webext',
-  'build:safari'
+  'build:safari',
```
andreasknoepfle: couldn't we instead change the eslint rule for the last entry ? 
Personally I find it strange that there is a `,` at the end. But thats only my Ruby-brain speaking.

spec/popup/components/header-spec.jsx
```diff
@@ -11,6 +11,7 @@ describe('header', () => {
 
   it('renders a link to the about page', () => {
     const wrapper = shallow(<Header tickets={tickets} />);
+    // eslint-disable-next-line jsx-a11y/anchor-is-valid
```
andreasknoepfle: seems strange to me. why do we need this disable?

gulpfile.babel.js
pmeinhardt: Hi @andreasknoepfle, disabling dangling commas was basically the one change that we made over the Airbnb config in our previous, own Bitcrowd config, because it looked so alien to many. At the same time, there's [good reasons](https://github.com/airbnb/javascript#commas--dangling) to actually write the code that way (cleaner diffs and – arguably – easier editing of array/object entries).

There's even languages – [like Go](https://play.golang.org/p/tm7yF0d5hb) – that require this as part of their syntax definition.

If it's not a problem, I'd keep it like this 😃

spec/popup/components/header-spec.jsx
pmeinhardt: To be honest I did not dig into this, because it was Friday evening. Seems a bit odd to me as well.

src/common/popup/components/external-link.jsx
pmeinhardt: I've added another commit to avoid the ESLint-rule override and explicitly required `children` as a prop 👋

planktonic: An anchor has the default role of `link`, adding it should be unnecessary. What’s this link linking to? I see a `target="_blank"` but no `href` — if this is only triggering some JS, perhaps it should be a `<button>`

pmeinhardt: Thanks for your input @planktonic. The handler is only there for Safari, which did not allow the use of regular anchors within extensions to open new browser tabs (see `openext` function that's passed down in the React context).

My thinking was: It's still semantically a link even if Safari makes you jump through hoops.

The commit I was referring to above actually adds the `href` and removes the ESLint override 😉

spec/popup/components/header-spec.jsx
pmeinhardt: I will try to take a look later today or tomorrow morning, so we can move on with this.

gulpfile.babel.js
andreasknoepfle: Fine for me 👍  
(But currently not so fine for the CI :P)

pmeinhardt: WAT? But only for the PR job? I'll take a look

pmeinhardt:
@andreasknoepfle the issues you pointed out should now be fixed. Feel free to take another look and squash+merge if you deem it good 😉

spec/popup/components/header-spec.jsx
andreasknoepfle: Did you find out why this is needed here?
Tempted to merge this right away, but can't let this be uncommented :P

pmeinhardt: Ah, foo! I missed that one. Here's the answer, will provide a fix ASAP.

pmeinhardt: https://github.com/evcohen/eslint-plugin-jsx-a11y/issues/339

pmeinhardt: Done 👌

Issue #64, opened 2017-10-15
Warnings when installing with a recent chrome


tessi:
Installing tickety-tick on a recent chrome throws a warning since PR #61 (which is supposed to make tickety-tick be installable on firefox again):

``` diff
diff --git a/src/web-extension/manifest.json b/src/web-extension/manifest.json
index 7b48559..4deca83 100644
--- a/src/web-extension/manifest.json
+++ b/src/web-extension/manifest.json
@@ -4,11 +4,6 @@
   "version": "#{VERSION}",
   "description": "#{DESCRIPTION}",
   "icons": {},
+  "applications": {
+    "gecko": {
+      "id": "tickety-tick@bitcrowd.net"
+    }
+  },
   "background": {
     "scripts": ["background.js"]
   },
```

The error ist:

```
There were warnings when trying to install this extension:
    Unrecognized manifest key 'applications'.
```

This was tested under chrome `Version 61.0.3163.100 (Official Build) (64-bit)` on MacOS.

Reverting this aforementioned commit "fixes" the issue (and probably removes firefox support again).

The extension installs even though there is that warning, still we might want to think about a way to get rid of it.

andreasknoepfle:
I researched a bit and it seems that one needs the `applications` tag in firefox only for manually installing versions from XPI files. 
But as it is probably still going to take a while until we can have a real mozilla app store release, we need this desperately to support newer Versions of Firefox.

https://stackoverflow.com/questions/37127330/unrecognized-manifest-key-applications-warning-for-google-chrome#37132274
https://developer.mozilla.org/en-US/Add-ons/WebExtensions/WebExtensions_and_the_Add-on_ID#When_do_you_need_an_Add-on_ID

Common solutions are apparently building a separate version for Firefox manual distribution.

andreasknoepfle:
![screen shot 2017-10-23 at 15 02 20](https://user-images.githubusercontent.com/239398/31890419-55939f82-b803-11e7-9bda-a19302cefdb9.png)

Seems like this is also blocking firefox release...

pmeinhardt:
@tessi & @andreasknoepfle, does this warning still occur?

I think it should be fixed by https://github.com/bitcrowd/tickety-tick/pull/65 (and in the more recent [#71](https://github.com/bitcrowd/tickety-tick/pull/71/files#diff-facb6a6e1a9f17b4ded04b4287bdba55R31)) plus we have a new Firefox release now.

[...]

PR #65, opened 2017-10-24
[#64] Warnings when installing with a recent chrome

andreasknoepfle:
Since the addon-ID is only needed when manually installing the plugin in newer versions
of firefox and also it is only needed if you want to install it permanently I added an env variable to activate the injection of the 'fake' addon ID. 

I put some references to #65

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,14 +67,14 @@ Navigate to the [chrome://extensions](chrome://extensions) page, enable
 
 ### Firefox
 
-For Firefox >48: Build the extensions with `WITH_ADDON_ID=true yarn build`
+For Firefox >=48: Build the extensions with `WITH_GECKO_ID=true yarn build`
 
 If you just want to try out and debug the extension, go to
 [about:debugging#addons](about:debugging#addons). Then press "Load Temporary
 Add-On" and select the `manifest.json` from the built extension directory.
 
-If you want to install this addon permanently goto [about:addons](about:addons)
-and klick on the small cogs icon. Select `Install Addon from file...` and choose
+If you want to install this addon permanently go to [about:addons](about:addons)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -67,10 +67,16 @@ Navigate to the [chrome://extensions](chrome://extensions) page, enable
...
 ### Firefox
 
+For Firefox >48: Build the extensions with `WITH_ADDON_ID=true yarn build`
+
 If you just want to try out and debug the extension, go to
 [about:debugging#addons](about:debugging#addons). Then press "Load Temporary
 Add-On" and select the `manifest.json` from the built extension directory.
 
+If you want to install this addon permanently goto [about:addons](about:addons)
+and klick on the small cogs icon. Select `Install Addon from file...` and choose
```
pmeinhardt: A few smaller typos:

1. goto => go to
2. klick => click
3. cogs => cog
4. Install Addon from file... => Install Add-on From File...

README.md
```diff
@@ -67,10 +67,16 @@ Navigate to the [chrome://extensions](chrome://extensions) page, enable
 
 ### Firefox
 
+For Firefox >48: Build the extensions with `WITH_ADDON_ID=true yarn build`
```
pmeinhardt: Just to make sure: Is it >48 or ≥48 (a.k.a. 48+)?

pmeinhardt: Also, how about naming the variable `WITH_GECKO_ID` which is less generic but closer to what actually happens under the hood?

andreasknoepfle: 👍

README.md
andreasknoepfle: Me != English

pmeinhardt:
Bueno! Go for it :shipit:

PR #66, opened 2017-10-24
release 1.2.2

andreasknoepfle:
Since we worked quite a lot on tickety-tick, I would try a new release also to finally achive 
#50.
Since the Firefox release seems to me to be the most crucial and hardest one, I will try and start with that one. 
before we can release however we need: 
- [x] #65 merged. 

In order to be in sync with the version numbers, I'd like to tick the number to 1.2.2, so we can later also release on chrome/opera.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "private": true,
-  "version": "1.2.1",
+  "version": "1.2.2",
   "name": "tickety-tick",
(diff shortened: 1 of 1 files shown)


pmeinhardt:
I added a few thoughts to #65. Thank you so much for taking on the Firefox release! 🤘

PR #71, opened 2017-10-31
Improve development experience

pmeinhardt:
- [x] Simplify build setup
- [x] Faster rebuilds (watch mode)
- [x] Add live-reloading
- [x] Add scripts to automate development tasks
- [x] Update readme


bitboxer:
Maybe [web-ext](https://github.com/mozilla/web-ext) can help.

pmeinhardt:
Yup, that's what I am looking at – for Firefox at least. But then there's a few more things. I was also thinking about moving away from Gulp, but haven't made up my mind yet.

I wonder why Travis is behaving weirdly…

image

andreasknoepfle:
Seems like web-ext can at least package Chrome extensions as well https://github.com/mozilla/web-ext/issues/119

As long as they only use a firefox compatible manifest...

src/common/popup/popup.scss
```diff
@@ -2,26 +2,26 @@
...
-@import "bootstrap/scss/nav";
-@import "bootstrap/scss/navbar";
-@import "bootstrap/scss/list-group";
+@import "~bootstrap/scss/nav";
+@import "~bootstrap/scss/navbar";
+@import "~bootstrap/scss/list-group";
 
 // Utility classes
-@import "bootstrap/scss/utilities";
+@import "~bootstrap/scss/utilities";
```
andreasknoepfle: out of curiosity. Why do we need `~` now to make it work?

pmeinhardt: https://github.com/webpack-contrib/sass-loader#imports

andreasknoepfle: 👍  thanks for the heads up. good to know

pmeinhardt:
Not exactly sure why Travis started failing with:

```
/home/travis/build/bitcrowd/tickety-tick/node_modules/eslint-config-airbnb-base/rules/best-practices.js:
	Configuration for rule "no-else-return" is invalid:
	Value "[object Object]" should NOT have more than 0 items.
```

pmeinhardt:
It appears `web-ext` depends on an outdated version of `eslint` and the older `yarn` that the Travis CI worker used, also used that one when running our `lint` task. However this eslint version (`4.8.0`) could not deal with `eslint-config-airbnb`'s ``"no-else-return"` rule definition.

Updating `yarn` in the Travis job fixed the issue (`before_install`).

pmeinhardt:
@andreasknoepfle, I'd be happy about another review/test-drive and your feedback 😃 

If everything's good, feel free to merge. This doesn't change the extension itself, so we do not need to publish a new release — it's solely changing how we build the extension, esp. an improved watch mode, and a couple of additional scripts to facilitate development.

See the updates to the [`README.md`](https://github.com/bitcrowd/tickety-tick/pull/71/files#diff-04c6e90faac2675aa89e2176d2eec7d8) for a bit more detail on the new `build:[browser]`, `watch:[browser]` and `open:[browser]` tasks.

pmeinhardt:
@bitboxer, we're using the `web-ext` command now to launch a Firefox instance with the plugin installed in a new temporary profile by the way 😉 Thanks for the hint!

.travis.yml
```diff
@@ -1,6 +1,10 @@
 language: "node_js"
-node_js: "6"
+node_js: "8"
 cache:
   directories:
     - "node_modules"
```
andreasknoepfle: we could do
``` 
cache: 
  yarn: true
``` 
now I think

andreasknoepfle: https://yarnpkg.com/lang/en/docs/install-ci/#travis-tab

webpack.webext.babel.js
```diff
@@ -0,0 +1,44 @@
...
+
+      mf.name = pkg.name;
+      mf.version = pkg.version;
+      mf.description = pkg.description;
+
+      if (variant === 'firefox') {
+        mf.applications = {
+          gecko: {
+            id: 'tickety-tick@bitcrowd.net',
+          },
```
andreasknoepfle: we only need the addon id on firefox if we want to manually install it. 
I think it could cause trouble in the version that we release in the store (if I remember correctly)

script/open-in-chrome
```diff
@@ -0,0 +1,21 @@
...
+  chromeFlags: [`--load-extension=${dir}`],
+  enableExtensions: true,
+  startingUrl: url,
+};
+
+launcher.launch(options).then(chrome => {
+  console.log(`Chrome debugging port on ${chrome.port}`);
+});
+
+// vim: set filetype=javascript
```
andreasknoepfle: nice ❤️

webpack.webext.babel.js
```diff
@@ -0,0 +1,44 @@
...
+        };
+      }
+
+      return JSON.stringify(mf);
+    },
+  },
+]]);
+
+export default config.toConfig();
```
andreasknoepfle: this is so nice ❤️ 👍

andreasknoepfle:
Thanks for your huge efforts @pmeinhardt . Great improvements! 
Just made two tiny comments :)

webpack.webext.babel.js
pmeinhardt: Aaah. Now I am confused 😕 😁 

I remember that in #54, we added the "applications" entry because there were issues when trying to install a locally built version in the Firefox Quantum preview.

However, this [Mozilla page](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/WebExtensions_and_the_Add-on_ID#When_do_you_need_an_Add-on_ID) says we do not need it all and I just tried it with the official Firefox Quantum 57.0 and it worked without complaints.

Can you please give it a try with these lines commented out @andreasknoepfle?

If successful, I'd say we remove those lines again.

pmeinhardt: …installed without gecko ID

andreasknoepfle: ### On Firefox:
You can: 
* Install it manually, temporarily with or without Addon-ID
* Install it manually, permanently with Addon-ID
* Install it from the marketplace, permanently without Addon-ID

There are errors when: 
* Installing manually, permanently without Addon-ID
* Releasing a version to the marketplace with a self-chosen Addon-ID 

### On Chrome:
You can:
* Do anything without an Addon-ID

Get warnings:
* with anything with an Addon-ID

andreasknoepfle: So the question is, do we want to keep the ability to manually install the addon, when firefox is not publishing it for quite a while again.

pmeinhardt: 🤦‍♂️ Oh, I see. Thanks for figuring that out… 👍

pmeinhardt: @andreasknoepfle: What do you think about separate `build:firefox` and `build:firefox-local` scripts where only the latter inserts the addon-ID?

andreasknoepfle: 👍, that would do the trick

pmeinhardt:
@andreasknoepfle, I've added separate scripts:

- `yarn build:firefox-local`
- `yarn watch:firefox-local`

which will add the `applications.gecko.id` to the generated `manifest.json`.

I've updated the README accordingly.

Let me know if somethings still missing, otherwise this could be merged.

webpack.webext.babel.js
```diff
@@ -0,0 +1,44 @@
+/* eslint-disable import/no-extraneous-dependencies */
+
+import config, { src, dist } from './webpack.common';
+import pkg from './package.json';
+
+const variant = process.env.VARIANT; // 'chrome', 'firefox' or 'opera'
```
andreasknoepfle: shouldn't this comment contain ``firefox-local`

andreasknoepfle: but I guess not so important.

pmeinhardt: Good point. Updated it and added a bit of additional explanation 👍

Issue #72, opened 2017-11-07
Remove jQuery dependency


pmeinhardt:
We're currently using only a small subset of jQuery for abstracting browser differences in the implementation of ticket system adapters (for Trello, Pivotal, JIRA, GitHub, Gitlab etc.).

Instead, we could implement the small number of adapter helpers we have in a cross-browser-compatible way — with modern APIs like [`element.querySelector`](https://developer.mozilla.org/en-US/docs/Web/API/Element/querySelector) — or at least use a lighter alternative to the full jQuery lib.

pmeinhardt:
This was addressed in #76. Closing 🎉

PR #76, opened 2017-11-18
[#72] Remove jQuery dependency

pmeinhardt:
Based on changes in #71. Therefore needs to be rebased after that is merged and for review, focus on the commits starting at `[#72] …`.

Removes jQuery as a dependency. Instead implements adapter helpers using native DOM APIs.

See the links for detailed information on browser support and references to the MDN docs et cetera.

- [`Element.querySelector`/`querySelectorAll`](https://caniuse.com/#search=queryselector)
- [`Element.dataset`](https://caniuse.com/#search=dataset)
- [`Element.classList`](https://caniuse.com/#search=classlist)
- [`Element.getAttribute`](https://caniuse.com/#search=getattribute)
- [`Element.closest`](https://caniuse.com/#search=closest) (with a shim for DOM implementations that do not support it)
- `Element.textContent`
- …

Unit tests are passing, but we need to make sure everything still works in the **supported browsers** (`jsdom` may behave differently, e.g. for `Element.closest`) and **story trackers**.

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,7 +63,6 @@
   },
   "dependencies": {
     "bootstrap": "4.0.0-alpha.4",
-    "jquery": "3.2.1",
     "octicons": "7.0.1",
     "prop-types": "15.6.0",
     "react": "16.1.1",
--- a/spec/support/helpers/setup.js
+++ b/spec/support/helpers/setup.js
@@ -2,12 +2,9 @@ import Enzyme from 'enzyme';
 import Adapter from 'enzyme-adapter-react-16';
 
 import { JSDOM } from 'jsdom';
-import jquery from 'jquery';
 
 Enzyme.configure({ adapter: new Adapter() });
 
 global.window = new JSDOM('<html><body></body></html>').window;
 global.document = global.window.document;
 global.navigator = global.window.navigator;
-
-global.$$$ = jquery(global.window);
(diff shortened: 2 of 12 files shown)


pmeinhardt:
@andreasknoepfle, @bitboxer I've just rebased and merged onto `master` (which now uses the updated build scripts with webpack). If you get a chance, to test-drive this branch/version which no longer includes jQuery I'd be happy about your feedback.

Maybe you can drop a line or two about which browser and ticket systems you've tried it on and whether you discovered any issues, because there might be subtle, breaking differences between the actual browsers and JSDOM which we use in our tests.

Cheers :wave

andreasknoepfle:
Firefox: 
- [x] jira
- [x] trello
- [x] github
- [x] gitlab
- [ ] pivotal (i don't have access to a existing pivotal project, maybe its ok to skip for now)

Chrome: 
- [x] jira
- [x] github
- [x] gitlab
- [x] trello
- [ ] pivotal

andreasknoepfle:
I'd say that is fine for now. If there is something off with pivotal or safari/opera we could also wait if there are bugtickets coming in, since safari has currently no store, opera takes forever anyway to publish (I think they don't review currently) and pivotal is currently hard to test.

pmeinhardt:
Awesome. Thanks a lot @andreasknoepfle for taking the time and testing these changes! 💚

Issue #75, opened 2017-11-16
Keyboard shortcuts


andreasknoepfle:
I think it would be a huge addition to support some basic keyboard shortcuts for tickety-tick.

Maybe something like:
* `ctrl-t` to open the tickety tick menu. (from there we can already navigate with the keyboard)
* `ctrl-alt-t` to get a commit message (???)  

There is a Webextensions API for this:
https://developer.mozilla.org/en-US/Add-ons/WebExtensions/manifest.json/commands

I'd think it would be probably ok if we skip Safari support here :P. Chrome and Firefox do support this. Not sure about Opera though.

andreasknoepfle:
With the basic shortcut we can already do most things, since we can tab trough and enter after. So I close this for now. Feel free to reopen when you have further suggestions, what shortcuts we really need.

PR #78, opened 2017-12-05
[#75] Keyboard shortcuts - basic

andreasknoepfle:
A basic shortcut is actually not that hard. 
For more shortcuts we need to handle them correctly, but for now this does most of the trick.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,12 @@ Tickety-Tick is available for every major browser:
 - [Opera](https://addons.opera.com/extensions/details/tickety-tick/)
 - For Safari, you need to build it yourself (see below)
 
+## Keyboard Shortcuts
+
+| Shortcut | Description                      |
+|----------|----------------------------------|
+| Ctrl + T | Open the extensions popup dialog |
+
 ## Building
 
 In order to build the extension from source, run:
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ Tickety-Tick is available for every major browser:
 
 | Shortcut | Description                      |
 |----------|----------------------------------|
-| Ctrl + T | Open the extensions popup dialog |
+| Ctrl + T | Open the extension's popup dialog |
 
 ## Building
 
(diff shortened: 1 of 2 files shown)


andreasknoepfle:
Probably need to update the docs accordingly :).

pmeinhardt:
Here's a link to help actually understand these changes:
https://developer.mozilla.org/en-US/Add-ons/WebExtensions/manifest.json/commands

src/web-extension/manifest.json
```diff
@@ -34,5 +34,15 @@
     "default_icon": "icon-128.png",
     "default_popup": "popup.html"
   },
+  "commands": {
+    "_execute_browser_action": {
+      "suggested_key": {
+        "default": "Ctrl+T",
+        "mac": "MacCtrl+T",
+        "chromeos": "Ctrl+T",
+        "linux": "Ctrl+T"
```
pmeinhardt: Do we actually need `"chromeos"` and `"linux"` because they're the same as the default?

At least according to the [docs](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/manifest.json/commands):
> The value for `"default"` is used on all platforms that are not explicitly listed.

pmeinhardt: A note in our `README.md` would be nice indeed.
Otherwise users might never find out about this feature.

andreasknoepfle: 👍 on both. Was more of a quick hack to try it out.

README.md
```diff
@@ -44,6 +44,12 @@ Tickety-Tick is available for every major browser:
 - [Opera](https://addons.opera.com/extensions/details/tickety-tick/)
 - For Safari, you need to build it yourself (see below)
 
+## Keyboard Shortcuts
+
+| Shortcut | Description                      |
+|----------|----------------------------------|
+| Ctrl + T | Open the extensions popup dialog |
```
pmeinhardt: Ah, this should be `… the extension's …` 😉

Issue #79, opened 2017-12-06
Release 1.3


andreasknoepfle:
Since we changed the infrastructure of the projects now by a lot we should do a new release soon.
With #77 & #78  we also have some features to roll out. What do you think @pmeinhardt? 
- [x] Mozilla Addon
- [x] Chrome extension
- [ ] Opera 

Lets see if Opera is even worth it, since it wasn't even considered by the admins for the 1.2.2 release #69.

andreasknoepfle:
I'm gonna close this, since everything was released except for Opera, where we are still waiting for older submissions... I think no one is reviewing on their side.

PR #80, opened 2017-12-06
[#79] Release 1.3

andreasknoepfle:
While I was rerecording VCRs I thought I can quickly prepare the tickety-tick release :).

As we did have an auto-zip feature before with gulp and the files need to be uploaded as zip to the stores I also added that quickly. Webpack is awesome sometimes ❤️ .

### DIFF
--- a/webpack.webext.babel.js
+++ b/webpack.webext.babel.js
@@ -1,5 +1,8 @@
 /* eslint-disable import/no-extraneous-dependencies */
 
+import path from 'path';
+import ZipWebpackPlugin from 'zip-webpack-plugin';
+
 import config, { src, dist } from './webpack.common';
 import pkg from './package.json';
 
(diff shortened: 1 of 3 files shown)


webpack.webext.babel.js
```diff
@@ -44,4 +47,12 @@ config.plugin('copy').tap(([patterns]) => [[
   },
 ]]);
 
+config.plugin('zip')
+  .use(ZipWebpackPlugin, [
+    {
+      path: path.join(__dirname, 'dist'),
```
pmeinhardt: Ah, this is what the `dist` helper is for 😉 I'll create a quick, tiny follow-up PR

pmeinhardt: Argh, and maybe we want to run zip each time when running in watch mode, so a separate task might be better.

Issue #73, opened 2017-11-07
Remove Bootstrap (alpha) as a dependency


pmeinhardt:
A candidate for replacing bootstrap could be [tachyons](http://tachyons.io/).

pmeinhardt:
Or let's just upgrade to Bootstrap 4.0 now that it's out.
Switching now is probably not worth the hassle.

pmeinhardt:
Done in #102

PR #84, opened 2017-12-19
 [#73] Remove Bootstrap (alpha) as a dependency 

andreasknoepfle:
Since it may be too hard to remove bootstrap entirely, I decided updating it for now. Sadly there is still only the beta release :(

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "zip-webpack-plugin": "2.1.0"
   },
   "dependencies": {
-    "bootstrap": "4.0.0-alpha.4",
+    "bootstrap": "4.0.0-beta.2",
     "octicons": "7.0.1",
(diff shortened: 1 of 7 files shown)


andreasknoepfle:
screen shot 2017-12-19 at 18 25 46

pmeinhardt:
Mmh, the design looks a bit broken: The summary button and info link are in the wrong place and their colors seem off. Like we discussed earlier, I'm not sure it's a big advantage to upgrade bootstrap from alpha to beta at the moment, esp. because we're also just using a small subset of it anyways and I wonder whether they'll ever release a stable next version… 😁

andreasknoepfle:
True. Working on a tachyons implementation now ... Not so much effort after all.

Issue #90, opened 2018-04-13
No tickets found on JIRA backlog view


lwassermann:
TT finds no issues on the JIRA backlog view, not even if a single item is selected (see #28 for more discussions if none are selected). This might be cause by a recent overhaul of the JIRA UI.

The `Active Sprints` view still works as desired.

pmeinhardt:
If you want to @lwassermann, we could pair on this 😉 I think I'll have a little time today/this week.

tessi:
should be clode by #92

PR #92, opened 2018-10-02
[#90] No tickets found on JIRA backlog view

tessi:
Enables finding tickets on the jira backlog view again. Also renames this within jira.js from `sidebar` to `backlog` which makes things a little more clear IMO

### DIFF
--- a/src/common/adapters/jira.js
+++ b/src/common/adapters/jira.js
@@ -1,4 +1,4 @@
-import { $find, $has, $text, $attr } from './helpers';
+import { $all, $find, $has, $text, $attr } from './helpers';
 
 const TYPES = ['bug', 'chore'];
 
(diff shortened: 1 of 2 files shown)


src/common/adapters/jira.js
```diff
@@ -20,18 +20,22 @@ const adapter = {
   inspect(loc, doc, fn) {
     if (doc.body.id !== 'jira') return fn(null, null);
 
-    if ($has('.ghx-fieldname-issuekey a', doc)) { // JIRA sidebar
-      const id = $text('.ghx-fieldname-issuekey a', doc);
-      const title = $text('[data-field-id="summary"]', doc);
-      const type = normalizeType($attr(`[data-issue-key="${id}"] .ghx-type`, doc, 'title'));
+    if ($has('.ghx-backlog-column .ghx-backlog-card.ghx-selected', doc)) {
```
klappradla: Just a quick look at the DOM in a Jira backlog and there seems to be an additional `ghx-selected-primary` class which is only used **once** (vs. one can select multiple tickets which will all have the `ghx-selected` class).

How does the current approach handle multiple tickets being selected? I assume it just take the very first one 🤔

tessi: good question. I'll investiage ⚙️

tessi:
The jira adapter now returns all currently selected tickets on the jira backlog view

screen shot 2018-10-03 at 13 40 15

Issue #97, opened 2018-10-20
Fix icon loading on about view


pmeinhardt:
The Tickety-Tick icon on the about view seems to not be loading at the moment (in all supported browsers). The code for that does not seem to have changed in a long time so I wonder what is causing this issue.

<img width="340" alt="image" src="https://user-images.githubusercontent.com/706519/47253677-01ee6500-d446-11e8-9ea5-26ab5113bf9f.png">

pmeinhardt:
🤦‍♂️ … the icons are currently copied using the `copy-webpack-plugin` using `{ flatten: true }`.

The path used in the DOM node rendered by the component does not match the build output.

tessi:
This should be fixed now thanks to @pmeinhardt

PR #100, opened 2018-10-20
[#97] Fix icon loading on about view

pmeinhardt:
Remove the icon as a simple way out.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "eslint-plugin-import": "2.14.0",
     "eslint-plugin-jsx-a11y": "6.1.2",
     "eslint-plugin-react": "7.11.1",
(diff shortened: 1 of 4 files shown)


tessi:
Hmmm can we inline the icon as a data-url instead?

pmeinhardt:
Absolutely @tessi. This was my initial thought. But then I realized I'll need to introduce another webpack loader and configure it and thought I'd see if I can get by without that 😁

I'll inline it right away if you prefer 🙂

pmeinhardt:
Fixed it now @tessi — I'll just have another look at the configuration and whether we could use this to explicitly import icons for generating the manifest etc.

image

pmeinhardt:
This is good to go from my side.

PR #105, opened 2018-10-25
Fix type in build command

klappradla:
Fix a typo in the package.json passing an invalid option "productionp" instead of "production" as the mode to build the Firefox release 👋


klappradla:
I'll quickly undo the new tag and the release commit to `master` - looks weird to release and then have the build script fixed in a later commit 😹

klappradla:
Thanx 💚

PR #109, opened 2018-10-31
Update docs for Opera

klappradla:
For the sake of consistency, I updated the README to include the same detail of information for Opera as for the other browsers:
- instructions on how to install a local add-on
- instructions and a screenshot on how to find the preferences


klappradla:
**-- UPDATE--**

I updated the README to no longer link to the non-existing page in the Opera store but instead to our own installation instructions in `OPERA.md`.

What do you think?

PR #114, opened 2018-12-04
Pin transitive dependency on event-stream

lwassermann:
by updating ps-tree

This was the smallest change I could find to achieve that.


klappradla:
I fear I do not really understand how this comes into play with the actual browser add-on 🙈 

Do we need to release a new version for this? Or does this only impact people _developing_ the extension?

PR #115, opened 2018-12-05
[TT-113] Does not work for the new JIRA version

### LINKED ISSUE #113: Does not work for the new JIRA version
JIRA changed their CSS identifiers to be obfuscated. (the page looks the same). This results in none of the JIRA pages working with tickety-tick anymore.

screenshot 2018-12-03 at 09 57 30

> ☝️ the HTML of a ticket in the backlog-view

We cannot use the (now obfuscated) css classes anymore and must rely on other means of scanning the page. Possible options are probably different per JIRA page. I looked at the backlog view a little closer and found these possible options:

screenshot 2018-12-03 at 10 02 15

1. `$('[data-react-beautiful-dnd-draggable="0"]')` to find tickets using the fact that they are draggable and JIRA uses this react library. The downside is that maybe non-tickets are draggable too and JIRA may change how they implement drag&drop.
1. search for `BFQT-38`-style links (ticket links) and remove those ticket-links which seem to be non-tickets. ` $('a').toArray().filter((a) => a.href.match(/\/browse\/.+-\d+/))` (finds all links which go to `/browse/BFQT-38`-like URLS). This should be fairly stable since all ticket-summaries contain a link to the actual ticket
1. search for the little bug/story/epic image in ticket-summaries. It's always an image of the form
  ```
  <img class="obfuscated" role="presentation" src="https://org.atlassian.net/secure/viewavatar?size=xsmall&amp;avatarId=11111&amp;avatarType=issuetype">
  ```
  So we can search for images with `avatarType=issuetype` in the URL

lwassermann:
- Try out old parsing before the new parsing, because it is better defined
- If none of the old methods worked, try
  - the title from the first h1 if the URL matches a single ticket url
  - to find links to browse and a title close by (backlog)
  - a dialog with a title and a link to a ticket, the first is assumed to be the jira ticket link

This would resolve #113 

*Notes:*
- The diff gets better if you ignore whitespace changes
- These changes a strongly bound to the assumption, that the Issue URL does not change
- It feels rather brittle, because positioning plays a strong role. Since I only have access to one new board, and can not change around in that, any setting my break the backlog functionality
- The `try {} catch () {}` also spans the callback. If there is an error in the callback, it will be called twice. That is not so nice, but at least the system will not explode?

*Notes 2:*
- Why is there a custom trim implementation, why not `String.prototype.trim`? Do we sometimes try to trim `null`? 
- JS DOM does not (and will not) support `innerText`. Instead of raising, it just returned `undefined`. 😒
- My impulse was to factor the adaptor implementation into many small functions. This current implementation tries to keep with the current style 🤷‍♂️

### DIFF
--- a/spec/adapters/jira-spec.js
+++ b/spec/adapters/jira-spec.js
@@ -2,237 +2,371 @@ import { JSDOM } from 'jsdom';
 
 import adapter from '../../src/common/adapters/jira';
 
-// parts of the dom of the jira backlog issue-list
-// contains two tickets - one of them being selected
-const BACKLOG = `
-  <div class="ghx-backlog-column">
-    <div class="ghx-backlog-card ghx-selected">
-      <div class="ghx-issue-content">
-        <div class="ghx-row ghx-plan-main-fields">
-          <span class="ghx-backlog-card-expander-spacer"></span>
-          <span class="ghx-type items-spacer" title="Story">
-            <img src="https://someorg.atlassian.net/secure/viewavatar?size=xsmall&amp;avatarId=12345&amp;avatarType=issuetype">
-          </span>
-          <div class="ghx-summary" data-tooltip="A Random JIRA Backlog Issue">
-            <span class="ghx-inner">A Random JIRA Backlog Issue</span>
+describe('old layout', () => {
+  // parts of the dom of the jira backlog issue-list
+  // contains two tickets - one of them being selected
+  const BACKLOG = `
+    <div class="ghx-backlog-column">
+      <div class="ghx-backlog-card ghx-selected">
+        <div class="ghx-issue-content">
+          <div class="ghx-row ghx-plan-main-fields">
+            <span class="ghx-backlog-card-expander-spacer"></span>
+            <span class="ghx-type items-spacer" title="Story">
+              <img src="https://someorg.atlassian.net/secure/viewavatar?size=xsmall&amp;avatarId=12345&amp;avatarType=issuetype">
+            </span>
+            <div class="ghx-summary" data-tooltip="A Random JIRA Backlog Issue">
+              <span class="ghx-inner">A Random JIRA Backlog Issue</span>
+            </div>
+          </div>
+          <div class="ghx-row ghx-end ghx-items-container">
+            <span class="aui-lozenge ghx-label ghx-label-single ghx-label-3" title="Chores" data-epickey="UXPL-123">
+              Chores
+            </span>
+            <span class="ghx-end ghx-items-container">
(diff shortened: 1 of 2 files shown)


spec/adapters/jira-spec.js
```diff
@@ -2,237 +2,371 @@ import { JSDOM } from 'jsdom';
...
+      </div>
+    </div>
+  </div>
+  `;
+
+  const BOARD_STORY = `
+  <div role="dialog">
+    <a href="/browse/UXPL-39"><span>UXPL-39</span></a>
+    <h1>A Random JIRA Board Issue</h1>
+  </div>
```
lwassermann: Because the HTML is _really_ complex, this is the minimal example that can be read by the implementation. This cuts ~20 levels of divs, and unused interactions.

spec/adapters/jira-spec.js
```diff
@@ -2,237 +2,371 @@ import { JSDOM } from 'jsdom';
...
+                   </span>
+                </span>
+                Create issue
+             </button>
+          </div>
+       </div>
+    </div>
+  `;
+
+  const STORYPAGE = `
```
lwassermann: I could not click-reproduce the case where the title is edited. In the new JIRA implementation, once the title looses focus, it is submitted. Was that the case before?

pmeinhardt:
@tessi, I think you're currently working on a project using Jira? Maybe you can give @lwassermann's changes a try 😉

tessi:
talked to lars about this. we find DOM parsing still a little brittle and want to experiment with fetching the ticket id and then doing a JQL query to the JIRA api to receive structured ticket details (including ticket descriptions etc. for a later expansion of tickety tick).

also we coulnd't test this on firefox. not sure why :/

pmeinhardt:
Cool. Thought about fetching from an API before but then never got around to doing it 👍

pmeinhardt:
FYI @lwassermann, @tessi: I ran into an issue when responding to the background script asynchronously from the content script. I need to investigate exactly why this is happening. This could mean that fetching ticket information through API requests won't work.

You can find some additional details in this PR: https://github.com/bitcrowd/tickety-tick/pull/129

I'll have a look to see whether/how this can be solved 🔎

pmeinhardt:
Alright. I've solved the issue and I am updating the adapter API to use async functions/promises. This way, we're using the APIs more like standardized by the W3C and it's clear that network requests for fetching ticket information will work as expected. 🙌

pmeinhardt:
We have a new candidate for a solution to this problem, #137, which uses the approach you @lwassermann and @tessi came up with. I'll close this one here now 🤗

PR #117, opened 2018-12-23
Use Jest

pmeinhardt:
Switch from Jasmine to Jest

➕ Support for snapshot testing (added snapshot tests for `About` and `NoTickets` components)
➕ Easier test coverage report generation/setup (see `README.md` changes)
➕ Test "watch" & "focus" modes are nice for development

➖ Slower tests (around 5s compared to only 2s with Jasmine)

Moved the test files from their separate `spec` directory right next to the implementation files.


README.md
```diff
@@ -127,18 +127,30 @@ To execute the automated source code checks, run:
 yarn checks
 ```
 
+Or, to run checks individually:
```
klappradla: 👍

src/common/popup/components/no-tickets.test.jsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+
+import NoTickets from './no-tickets';
+
+describe('about', () => {
```
klappradla: Some copy paste leftover?

src/common/popup/components/about.test.jsx
```diff
@@ -0,0 +1,14 @@
...
+
+import About from './about';
+
+describe('about', () => {
+  function render(/* overrides */) {
+    return shallow(<About />);
+  }
+
+  it('renders', () => {
+    expect(render({})).toMatchSnapshot();
```
lwassermann: Have you thought about using inline snapshots (`toMatchInlineSnapshot`)? I see that the snapshots are a bit longer, but that would add some more information into the test file.

pmeinhardt: 🤔 I have not. I will give it a shot and see what that looks like 👍 Great suggestion.

pmeinhardt: Mmh, looks a bit odd because the whole snapshot is not indented properly relative to the surrounding code… and it's really quite long. Maybe keeping it in a separate file is the easier option? What do you think @lwassermann?

PS: Since inline snapshots require prettier though and I've just set it up for trying it out, I'll open a new PR for the prettier integration.

PR #129, opened 2018-12-23
Update adapter code

pmeinhardt:
**Description**

Make adapters async functions so they're more intuitive, e.g. with regard to fetching ticket info from a web API. Previously they used callback style with error and ticket arguments for the callback. However the error argument was never actually used/set.

Now, adapters should export an async `scan` function that receives the page `location` and `document` and returns an array of tickets.

Example: [github.js](https://github.com/bitcrowd/tickety-tick/blob/update-adapter-code/src/common/adapters/github.js)
Compared to the previous way of writing adapters: [github.js (old)](https://github.com/bitcrowd/tickety-tick/blob/871a8a34fb1581dbd579750fa57023be0888a2a8/src/common/adapters/github.js)

Hint: Enable "Hide whitespace changes" in the diff options 😉 

---

Intermediate issue (for the record only):

Both Chrome and Firefox seem to have issues with asynchronous code when communicating between different extension scripts. In particular, the content script is unable to send a response to the background page when sending from a promise/timeout callback ([see code comment](https://github.com/bitcrowd/tickety-tick/commit/ad6ea6b85699870070f23fd8268bc37a90cb3ac8#diff-92ed3a44e047c3ddece099f21d73b7c5R10)).

🤦‍♂️ 

This needs some investigation.

debugging

☝️ `port` is `null`

### DIFF
--- a/src/common/adapters/github.js
+++ b/src/common/adapters/github.js
@@ -5,49 +5,47 @@ import {
   $value,
 } from './helpers';
 
-const adapter = {
-  inspect(loc, doc, fn) {
-    if ($has('.issues-listing .js-issue-row.selected', doc)) {
-      const issues = $all('.issues-listing .js-issue-row.selected', doc);
+async function scan(loc, doc) {
+  if ($has('.issues-listing .js-issue-row.selected', doc)) {
+    const issues = $all('.issues-listing .js-issue-row.selected', doc);
 
-      const tickets = issues.map((issue) => {
-        const id = $value('input.js-issues-list-check', issue);
-        const title = $text('a.js-navigation-open', issue);
-        const labels = $all('.labels .label', issue);
-        const type = labels.some(l => /bug/i.test(l.textContent)) ? 'bug' : 'feature';
+    const tickets = issues.map((issue) => {
+      const id = $value('input.js-issues-list-check', issue);
+      const title = $text('a.js-navigation-open', issue);
+      const labels = $all('.labels .label', issue);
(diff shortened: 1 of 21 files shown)


src/common/adapters/github.js
```diff
@@ -5,49 +5,47 @@ import {
...
+      return { id, title, type };
+    });
+
+    return tickets;
+  }
+
+  return null;
+}
+
+export default scan;
```
pmeinhardt: It's just the indentation and the return values that have changed here (we no longer `return fn(null, tickets)`, but just `tickets` now that we're using an async function).

pmeinhardt: Same for the other adapters.

src/common/search.js
```diff
@@ -7,25 +7,9 @@ import Trello from './adapters/trello';
...
-
-      results[i] = err ? null : res;
-      pending -= 1;
-
-      if (pending === 0) done();
-    });
-  });
+export async function search(adapters, loc, doc) {
+  const results = await Promise.all(adapters.map(scan => scan(loc, doc)));
+  return results.find(e => (e !== null)) || null;
```
lwassermann: 👍 
This is much clearer to me.

src/web-extension/options/options.jsx
```diff
@@ -1,5 +1,3 @@
-/* eslint-env browser */
```
lwassermann: 👍

PR #130, opened 2019-01-07
Use WebExtension polyfill to avoid browser compatibility issues

pmeinhardt:
Adds [`mozilla/webextension-polyfill`](https://github.com/mozilla/webextension-polyfill).

This allows us to use the soon-to-be-standardized (according to the polyfill documentation and MDN), promise-based APIs across browsers.

We are using, for instance, [`storage.get()`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/storage/StorageArea/get) and [`storage.set()`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/storage/StorageArea/set).

Additionally, switching to the promise-based APIs also seems like an improvement, because we've updated our adapter code to use `async` & `await` and we now handle other asynchronous behavior in the same, consistent way.

Another small but nice side-effect is that we can drop some linting exceptions.

### DIFF
--- a/src/web-extension/options/components/form.test.jsx
+++ b/src/web-extension/options/components/form.test.jsx
@@ -7,68 +7,73 @@ import TemplateInput from './template-input';
 import { defaults as fallbacks, helpers } from '../../../common/format';
 
 describe('form', () => {
-  function render(overrides) {
-    const store = { get: jest.fn(), set: jest.fn() };
+  function render(overrides, options = {}) {
+    const store = { get: jest.fn().mockResolvedValue({}), set: jest.fn().mockResolvedValue({}) };
     const defaults = { store };
 
     const props = { ...defaults, ...overrides };
-    const wrapper = shallow(<Form {...props} />);
-    const instance = wrapper.instance();
+    const wrapper = shallow(<Form {...props} />, options);
 
-    return { wrapper, instance, props };
+    return wrapper;
   }
 
   const inputs = wrapper => wrapper.find(TemplateInput);
   const input = (wrapper, name) => inputs(wrapper).find({ name });
   const value = (wrapper, name) => input(wrapper, name).prop('value');
 
-  const change = (wrapper, instance, name, val) => {
+  const change = (wrapper, name, val) => {
     const event = { target: { name, value: val } };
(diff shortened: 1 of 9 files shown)


src/web-extension/options/components/form.test.jsx
```diff
@@ -7,145 +7,151 @@ import TemplateInput from './template-input';
...
+
+    expect(field.props()).toEqual({
       label: 'Branch Name Format',
       id: 'branch-name-format',
       name: 'branch',
       value: '',
       fallback: fallbacks.branch,
       disabled: true,
-      onChange: instance.handleChanged,
+      onChange: expect.any(Function),
```
lwassermann: You could also use `wrapper.instance().handleChanged` if you want to be explicit.

pmeinhardt: Good point. My idea with updating the tests was to make them less implementation-dependent and focus more on the effects of user interactions. At the same time it probably doesn't hurt to be more specific here. I'll update the test 🙇

pmeinhardt: Updated the tests. Thanks @lwassermann :metal

src/web-extension/options/components/form.test.jsx
```diff
@@ -68,84 +79,85 @@ describe('form', () => {
...
+    let saved;
+
+    const unchanged = { branch: 'branch', commit: 'commit', command: 'command' };
+    store.get.mockResolvedValue({ templates: unchanged });
+    store.set.mockReturnValue(new Promise((resolve) => {
+      // Create a new async function that resolves the promise returned from
+      // calls to `store.set()` and returns a promise itself which we can wait
+      // on to ensure data saving has been handled.
+      saved = async function save() { resolve(); };
+    }));
```
pmeinhardt: Not sure if this is easily understood or whether/how to improve it.

Basically I liked being able to say `await saved()` further down below to trigger "saving" and wait for all dependent promises (i.e. those re-enabling the form inputs) to finish processing.

@lwassermann, in case you get a chance to take another look, I'd be happy about your feedback 🙂

Issue #106, opened 2018-10-26
Temporarily load Firefox addon


klappradla:
I just realized it's not possible to temporarily load the Firefox build build via

    yarn build:firefox

The popup won't load any content:

<img width="1792" alt="screen shot 2018-10-26 at 12 25 01" src="https://user-images.githubusercontent.com/3491815/47561120-b3940700-d91a-11e8-82f1-d74a41f448e0.png">

This is also reported when submitting a new version to the Firefox addons place:

<img width="1792" alt="screen shot 2018-10-26 at 12 29 52" src="https://user-images.githubusercontent.com/3491815/47561187-da523d80-d91a-11e8-9722-37ddb03fd7ad.png">

They essentially report these warnings:

```
"storage.sync" can cause issues when loaded temporarily
Warning: This API can cause issues when loaded temporarily using about:debugging in Firefox unless you specify applications > gecko > id in the manifest. Please see: https://mzl.la/2hizK4a for more.
options.js line 6 column 1237

"storage.local" can cause issues when loaded temporarily
Warning: This API can cause issues when loaded temporarily using about:debugging in Firefox unless you specify applications > gecko > id in the manifest. Please see: https://mzl.la/2hizK4a for more.
options.js line 6 column 1258

Unsafe assignment to innerHTML
Warning: Due to both security and performance concerns, this may not be set using dynamic values which have not been adequately sanitized. This can lead to security issues or fairly serious performance degradation.
options.js line 22 column 35324

Unsafe assignment to innerHTML
Warning: Due to both security and performance concerns, this may not be set using dynamic values which have not been adequately sanitized. This can lead to security issues or fairly serious performance degradation.
options.js line 22 column 35347

"storage.sync" can cause issues when loaded temporarily
Warning: This API can cause issues when loaded temporarily using about:debugging in Firefox unless you specify applications > gecko > id in the manifest. Please see: https://mzl.la/2hizK4a for more.
popup.js line 6 column 1237

"storage.local" can cause issues when loaded temporarily
Warning: This API can cause issues when loaded temporarily using about:debugging in Firefox unless you specify applications > gecko > id in the manifest. Please see: https://mzl.la/2hizK4a for more.
popup.js line 6 column 1258

Unsafe assignment to innerHTML
Warning: Due to both security and performance concerns, this may not be set using dynamic values which have not been adequately sanitized. This can lead to security issues or fairly serious performance degradation.
popup.js line 22 column 35324

[...]

PR #132, opened 2019-01-17
[#106] Temporarily load Firefox addon

pmeinhardt:
Partially addresses #106.

Note ⚠️ I don't think there's much we can do to get rid of the warning about `innerHTML` at the moment because it is used for Octicons and friends.

---

Always set an addon-id when building tickety-tick for Firefox.
It appears we should always do so since we are using `storage.sync`.

The relevant bits from [MDN](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/WebExtensions_and_the_Add-on_ID):

  > Extensions can explicitly set the add-on ID using the `applications`
    key in manifest.json. However, this key is usually optional.

  > Note, though, that some WebExtension APIs use the add-on ID and
    expect it to be the same from one browser session to the next. If
    you use these APIs in Firefox, then you must set the ID explicitly
    using the `applications` key.

  > Some WebExtension APIs use the add-on ID and expect it to be the
    same from one browser session to the next. If you use these APIs,
    then you must set the ID explicitly using the `applications` key.
    This applies to the following APIs: […], `storage.sync`, […]


pmeinhardt:
There is one thing I'd like to double check with you @klappradla ⚠️ 

Currently the webpack build sets the addon-id to `tickety-tick@bitcrowd.net`. However this was only used for `firefox-local` builds before, not the ones that we submit to Mozilla.

When I open `about:addons` and move on to "Debug add-ons", Firefox tells me the "Extension ID" for the extension installed from the official Mozilla site is `jid1-ynkvezs8Qn2TJA@jetpack`.

I would like to avoid issues with changing the addon-id, i.e. users loosing their stored preferences because of this (since we're using `storage.sync`/`storage.local` for these and they appear to use the addon-id).

Could you please check which "Extension ID" you see in Firefox? Is it the same as for you?

debug add-ons

If it is the same for you, then it might be safer for us to set this one in the manifest.

klappradla:
Looks like it's the same 👀 

screenshot 2019-01-18 at 00 43 42

pmeinhardt:
@klappradla, I've updated the manifest generation to use the official, random addon-id tickety-tick apparently received when we submitted it. I've checked locally with `yarn build:firefox-local` and `yarn open:firefox-local` and everything seems to work.

Curious to learn what Mozilla thinks of this, when you next submit it 🙂

pmeinhardt:
I just noticed we probably do not need the `firefox-local` variant any more if this works.

README.md
```diff
@@ -77,15 +77,9 @@ Navigate to the [chrome://extensions](//chrome://extensions) page, enable "Devel
 
 ### Firefox
 
-Build the extension with:
+If you just want to try out and debug the extension, go to [about:debugging#addons](//about:debugging#addons). Then press "Load Temporary Add-On" and select the built `manifest.json` from the `dist/firefox` extension directory.
```
klappradla: 🎉

Issue #93, opened 2018-10-19
Enhance commit messages


pmeinhardt:
Thanks to some inspiration from @klappradla, I would like to propose working on the following improvements:

1. [x] Extend ticket objects (currently `id`, `type`, `title`, `fmt`) by adding `description` and `url` if available
2. [x] Format commit messages with a 50 char title, empty line, commit description with a text width of 72 chars — https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

If you have additional ideas regarding better commit messages, let's gather them here 🙂 

cc @tessi, @maltoe

tessi:
A link, if possible to the original ticket would be great.

pmeinhardt:
Just two quick notes for future selves 💭

1. **Adding ticket URLs**

    My first thought was: Easy! We already pass the document location as an argument to the adapters.

    On second thought: The scanner probably needs to add the URL on a case-by-case basis. Consider the GitHub issues list for instance, where you can select multiple issues via their checkboxes. The ticket URLs shouldn't be `https://github.com/<orga>/<repo>/issues` but `…/issues/<id>`.

2. **Commit message formatting**

    I have been thinking about how to properly format commit messages (wrap title after 50 chars, text after 72 chars, with proper alignment/indents…) and prettier's markdown support could make it a valid option (for most of this):

    https://is.gd/DVf7Jt 💅 (uses `--text-width 72`, `--prose-wrap always`)

    I'll keep an eye out for more lightweight alternatives though 👀

pmeinhardt:
The auto-formatting is done. While a few adapters already return the ticket URL etc., most of them don't. While this additional information is optional, these adapters should still be updated ✌️

PR #133, opened 2019-01-18
[#93] Enhance commit messages  (part 1)

pmeinhardt:
_Edited_

As per https://github.com/bitcrowd/tickety-tick/pull/133#issuecomment-473600952, this could be the first part of a series of PRs to enhance commit messages generated by Tickety-Tick.

**Changes**

* Enhance the default commit message format to include ticket descriptions and URLs if present
* Update the options page to make editing of multi-line templates more convenient

**Preview**

preview

hints


klappradla:
Sorry I did not comment on this earlier 🙈 

Some cents 💸 

All in all: wow 🤓 ! Using `prettier` for the formatting is an awesome idea and the result also quite neat!

For the usefulness: I think it can work sort of like a shortcut to quickly going over the ticket description once again before finally committing all changes. One definitely has to go over the content again though (especially if it contains images, etc.)

I'm personally not using the *"pre-build"* commit messages that `TicketyTick` offers - simply because my usual workflow is different and I prefer to write the message from scratch. But if it comes with a bearable maintenance effort, offering the pre-filled message bodies can definitely be useful for certain people. It gives people a start for a commit message, one is not confronted with a blank page but can rather edit/adapt what's there.

However, if it comes with a huge maintenance effort and needs constant care on all adapters we support, I'd rather not do it but concentrate on the core things in `TicketyTick`: branch name and PR name.

tessi:
I like this PR, because (as max mentioned) it gives developers a commit-message body to start editing. We should (when the other PR is merged) however use the JIRA api to receive that data instead of page scraping.

pmeinhardt:
Okay. I've given this some more thought after taking a step back and came to the following conclusion:

**Normalization of ticket descriptions should be handled in the adapters**

My previous spike simply assumed that we can use prettier to consistently format ticket descriptions from all ticket tracking systems. However, each ticket tracking system uses a different markup language for ticket descriptions. E.g. GitHub uses Markdown, Jira uses their own markup syntax etc.

Thus, I think it is best to perform any normalization of ticket descriptions in our adapters. The formatting code can then expect a somewhat consistent textual representation for descriptions from different sources without parsing.

**This commit does not**

- Update any adapters to return ticket descriptions and/or URLs
- Perform any markup parsing/normalization in the adapters
- Format commit messages according to the recommendation (textwidth/wrapping…)

These will be implemented in follow-up PRs!

**What it does is**

- Enhance the default commit message format to include ticket descriptions and URLs if present
- Update the options page to make editing of multi-line templates more convenient

**Preview**

screencast

src/common/format/index.js
```diff
@@ -15,21 +15,22 @@ const fallbacks = {
...
-  const commit = renderer('commit');
-  const branch = renderer('branch');
-  const cmd = renderer('command');
+export default (templates = {}) => {
+  const branch = renderer(templates, 'branch');
+  const commit = renderer(templates, 'commit');
 
-  const command = values => cmd({
+  // eslint-disable-next-line no-underscore-dangle
+  const _command = renderer(templates, 'command');
```
andreasknoepfle: how about we call this `commandRenderer` so we do not need the `eslint-disable...`

src/web-extension/options/components/template-input.jsx
```diff
@@ -1,5 +1,18 @@
...
+  if (multiline) {
+    return <TextareaAutosize {...props} />;
+  }
+
+  return <input type="text" {...props} />;
+}
+
+TemplateInputElement.propTypes = {
+  multiline: PropTypes.bool.isRequired,
+};
```
andreasknoepfle: maybe we can put this into it's own file? Just an idea. It seems also not too terrible here though, so feel free to leave it.

src/common/format/index.js
pmeinhardt: 🤔 I used `cmd` before, but that seemed a bit cryptic. Using `commandRenderer` seems a bit odd too, because it is essentially the same kind of "thing" as the `branch` and `commit` values above, but these aren't called `*Renderer`. 🤷‍♂️ I honestly don't know… naming is hard. To me the current name didn't feel ideal, but the alternatives I could think of didn't either.

pmeinhardt:
Thanks a lot for your feedback @andreasknoepfle. I'm not really sure how to go about the naming…
I've added a hint on available template variables on the options page though, so people know what's there.

src/common/format/index.js
andreasknoepfle: Also fine with sticking with it :)

PR #137, opened 2019-03-04
Update Jira adapter to use REST APIs for fetching ticket info

nickgnd:
Update the JIRA adapter to fetch the ticket info via HTTP request instead of scraping the document.

Here's the documentation of the REST APIs: https://developer.atlassian.com/cloud/jira/software/rest/

I tested the changes both with Chrome and Firefox.

TODO:

- [x] Update adapter code to determine the proper URL base (including path prefix) instead of just using the hostname
- [x] Test on self-hosted Jira instance
- [ ] Test with Safari
- [ ] Test with Opera

Side notes:

- The package [`ky`](https://github.com/sindresorhus/ky) has been added and it is used as HTTP client

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -10,4 +10,5 @@ module.exports = {
     '^.+\\.jsx?$': 'babel-jest',
     '^.+\\.png$': '<rootDir>/test/transforms/file.js',
   },
+  transformIgnorePatterns: ['node_modules/(?!ky)'],
 };
--- a/package.json
+++ b/package.json
@@ -64,6 +64,8 @@
   "dependencies": {
     "@githubprimer/octicons-react": "^8.5.0",
     "bootstrap": "4.3.1",
+    "ky": "^0.9.0",
+    "micro-match": "^0.2.0",
     "prop-types": "15.7.2",
     "react": "16.8.4",
     "react-dom": "16.8.4",
--- a/src/common/adapters/jira.js
+++ b/src/common/adapters/jira.js
@@ -15,17 +15,25 @@ import client from '../client';
 
 function isJiraPage(loc, doc) {
   if (loc.host.endsWith('.atlassian.net')) return true;
-  if (doc.body.id === 'jira') return true;
+  if (doc.body.id === 'jira') return true; // self-managed instance on different domain
   return false;
 }
 
-function getSelectedIssueId({ href, pathname }) {
-  const { searchParams: params } = new URL(href);
+const pathSuffixes = new RegExp('/(browse/[^/]+|projects/[^/]+/issues/[^/]+|secure/RapidBoard.jspa)$', 'g');
+
+function getPathPrefix(loc) {
+  return loc.pathname.replace(pathSuffixes, '');
+}
+
+function getSelectedIssueId(loc, prefix = '') {
+  const { searchParams: params } = new URL(loc.href);
 
   if (params.has('selectedIssue')) return params.get('selectedIssue');
 
+  const path = loc.pathname.substr(prefix.length); // strip path prefix
+
   return ['/projects/:project/issues/:id', '/browse/:id']
-    .map(pattern => match(pattern, pathname).id)
+    .map(pattern => match(pattern, path).id)
     .find(Boolean);
 }
 
@@ -38,11 +46,12 @@ function extractTicketInfo(response) {
 async function scan(loc, doc) {
   if (!isJiraPage(loc, doc)) return null;
 
-  const id = getSelectedIssueId(loc);
+  const prefix = getPathPrefix(loc); // self-managed instances may host on a subpath
+  const id = getSelectedIssueId(loc, prefix);
 
   if (!id) return null;
 
-  const jira = client(`https://${loc.host}/rest/api/latest`);
+  const jira = client(`https://${loc.host}${prefix}/rest/api/latest`);
   const response = await jira.get(`issue/${id}`).json();
   const ticket = extractTicketInfo(response);
 
--- a/src/common/adapters/jira.js
+++ b/src/common/adapters/jira.js
@@ -1,61 +1,52 @@
-import {
-  $all,
-  $find,
-  $has,
-  $text,
-  $attr,
-} from './helpers';
+// Jira adapter
+//
+// This adapter extracts the identifier of the selected issue from the page URL
+// and uses the Jira API to retrieve the corresponding ticket information.
+//
+// https://developer.atlassian.com/server/jira/platform/rest-apis/
+//
+// * Backlog and Active Sprints: https://<YOUR-SUBDOMAIN>.atlassian.net/secure/RapidBoard.jspa?…&selectedIssue=<ISSUE-KEY>
+// * Issues and filters: https://<YOUR-SUBDOMAIN>.atlassian.net/projects/<PROJECT-KEY>/issues/<ISSUE-KEY>
+// * Issue view: https://<YOUR-SUBDOMAIN>.atlassian.net/browse/<ISSUE-KEY>
 
-const TYPES = ['bug', 'chore'];
+import match from 'micro-match';
 
-const normalizeType = (type) => {
-  const sanitizedType = type && type.toLowerCase();
-  if (TYPES.indexOf(sanitizedType) > -1) return sanitizedType;
-  return 'feature';
-};
+import client from '../client';
 
-const ticketPageTitle = (issue) => {
-  if ($has('#summary-form', issue)) { // ticket title is currently being edited
-    return $text('#summary-form #summary', issue);
-  }
+function isJiraPage(loc, doc) {
+  if (loc.host.endsWith('.atlassian.net')) return true;
+  if (doc.body.id === 'jira') return true;
+  return false;
+}
 
-  return $text('#summary-val', issue);
-};
+function getSelectedIssueId({ href, pathname }) {
+  const { searchParams: params } = new URL(href);
+
+  if (params.has('selectedIssue')) return params.get('selectedIssue');
+
+  return ['/projects/:project/issues/:id', '/browse/:id']
+    .map(pattern => match(pattern, pathname).id)
+    .find(Boolean);
+}
+
+function extractTicketInfo(response) {
+  const { key: id, fields: { issuetype, summary: title } } = response;
+  const type = issuetype.name.toLowerCase();
+  return { id, title, type };
+}
 
 async function scan(loc, doc) {
-  if (doc.body.id !== 'jira') return null;
+  if (!isJiraPage(loc, doc)) return null;
 
-  if ($has('.ghx-backlog-column .ghx-backlog-card.ghx-selected', doc)) {
-    // ticket list with backlog and sprints
-    const issueCssPath = '.ghx-backlog-column .ghx-backlog-card.ghx-selected';
-    const issues = $all(issueCssPath, doc).map((issue) => {
-      const id = $text('.ghx-key', issue);
-      const title = $text('.ghx-summary .ghx-inner', issue);
-      const type = normalizeType($attr('.ghx-type', issue, 'title'));
-      return { id, title, type };
-    });
-    return issues;
-  }
+  const id = getSelectedIssueId(loc);
 
-  if ($has('#issue-content', doc)) {
-    // ticket show-page, when a single ticket is opened full-screen
-    const issue = $find('#issue-content', doc);
-    const id = $text('#key-val', issue);
-    const title = ticketPageTitle(issue);
-    const type = normalizeType($text('#type-val', issue));
-    return [{ id, title, type }];
-  }
+  if (!id) return null;
 
-  if ($has('.ghx-columns .ghx-issue.ghx-selected', doc)) {
-    // board view, when a ticket is opened in a modal window
-    const issue = $find('.ghx-columns .ghx-issue.ghx-selected', doc);
-    const id = $attr('.ghx-key', issue, 'aria-label');
-    const title = $text('.ghx-summary', issue);
-    const type = normalizeType($attr('.ghx-field-icon', issue, 'data-tooltip'));
-    return [{ id, title, type }];
-  }
+  const jira = client(`https://${loc.host}/rest/api/latest`);
+  const response = await jira.get(`issue/${id}`).json();
+  const ticket = extractTicketInfo(response);
 
-  return null;
+  return [ticket];
 }
 
 export default scan;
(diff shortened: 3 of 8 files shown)


src/common/clients/request.test.js
```diff
@@ -0,0 +1,99 @@
...
+  it('rejects with an error on failure', () => {
+    const body = JSON.stringify({ status: 401, message: 'Unauthorized' });
+    get.mockImplementation(() => {
+      const response = new MockResponse(body, 401);
+      throw new HTTPError(response);
+    });
+
+    expect(() => request('https://test-error')).toThrowError(HTTPError);
+  });
+});
```
pmeinhardt: I think if we switch over to using the simplified method of creating clients with `ky.extend()` outlined above, we don't need these tests. Instead, we could just test that calling `client('https://example.io/api', { timeout: 1000 })` calls `ky.extend` with the correct options (defaults + overrides) and returns the resulting client:

```javascript
import ky from 'ky';

import client from './client';

jest.mock('ky', () => ({ extend: jest.fn() }));

describe('client', () => {
  const mock = { ky: true };

  beforeEach(() => {
    ky.extend.mockReturnValue(mock);
  });

  afterEach(() => {
    ky.extend.mockReset();
  });

  it('creates a new client with default values', () => {
    const base = 'https://example.io/api';
    const instance = client(base);
    expect(ky.extend.calls).toMatchSnapshot();
    expect(instance).toBe(mock);
  });

  it('sets custom options', () => {
    const instance = client('https://example.io/api', { credentials: 'same-origin', timeout: 1000 });
    expect(ky.extend.calls).toMatchSnapshot();
    expect(instance).toBe(mock);
  });
});
```

src/common/adapters/jira.js
```diff
@@ -1,61 +1,54 @@
...
-const ticketPageTitle = (issue) => {
-  if ($has('#summary-form', issue)) { // ticket title is currently being edited
-    return $text('#summary-form #summary', issue);
-  }
+// JIRA adapter
+// It gets the selected issue from the current URL and then it fetches the ticket info
+//
+// * Backlog and Active Sprints: https://<YOUR-SUBDOMAIN>.atlassian.net/secure/RapidBoard.jspa?…&selectedIssue=<ISSUE-KEY>
+// * Issues and filters: https://<YOUR-SUBDOMAIN>.atlassian.net/projects/<PROJECT-KEY>/issues/<ISSUE-KEY>
+// * Issue view: https://<YOUR-SUBDOMAIN>.atlassian.net/browse/<ISSUE-KEY>
```
pmeinhardt: I think this is super-useful 💚

src/common/adapters/jira.js
```diff
@@ -1,61 +1,54 @@
...
-  if (TYPES.indexOf(sanitizedType) > -1) return sanitizedType;
-  return 'feature';
-};
-
-const ticketPageTitle = (issue) => {
-  if ($has('#summary-form', issue)) { // ticket title is currently being edited
-    return $text('#summary-form #summary', issue);
-  }
+// JIRA adapter
+// It gets the selected issue from the current URL and then it fetches the ticket info
```
pmeinhardt: I would add an empty comment line above this one, update text just a wee bit and include a link to the API documentation:

```javascript
// Jira adapter
//
// This adapter extracts the identifier of the selected issue from the page URL
// and uses the Jira API to retrieve the corresponding ticket information.
//
// https://developer.atlassian.com/cloud/jira/software/rest/

File: src/common/adapters/jira.js [hunk 4]
```diff
@@ -1,61 +1,54 @@
...
-  const sanitizedType = type && type.toLowerCase();
-  if (TYPES.indexOf(sanitizedType) > -1) return sanitizedType;
-  return 'feature';
-};
-
-const ticketPageTitle = (issue) => {
-  if ($has('#summary-form', issue)) { // ticket title is currently being edited
-    return $text('#summary-form #summary', issue);
-  }
+// JIRA adapter
```
pmeinhardt: "JIRA" is actually spelled "Jira".

jest.config.js
```diff
@@ -10,4 +10,5 @@ module.exports = {
     '^.+\\.jsx?$': 'babel-jest',
     '^.+\\.png$': '<rootDir>/test/transforms/file.js',
   },
+  transformIgnorePatterns: ['node_modules/?!(ky)'],
```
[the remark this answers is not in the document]
nickgnd: 😨 🤦‍♂️ Opppsss...thanks for the catch!

package.json
```diff
@@ -66,7 +66,9 @@
   "dependencies": {
     "@githubprimer/octicons-react": "^8.2.0",
     "bootstrap": "4.2.1",
+    "ky": "^0.9.0",
     "prop-types": "15.6.2",
+    "qs": "^6.6.0",
```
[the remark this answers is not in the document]
nickgnd: Good to know, thanks 👍

nickgnd:
@pmeinhardt I updated the PR with your suggestions, thanks again for the feedback and tips, really really valuable ❤️

andreasknoepfle:
Question: Does this also work with self hosted/older versions of JIRA? If not, we might want the old strategy as a fallback, no?

src/common/adapters/jira.js
```diff
@@ -1,61 +1,57 @@
...
+// * Backlog and Active Sprints: https://<YOUR-SUBDOMAIN>.atlassian.net/secure/RapidBoard.jspa?…&selectedIssue=<ISSUE-KEY>
+// * Issues and filters: https://<YOUR-SUBDOMAIN>.atlassian.net/projects/<PROJECT-KEY>/issues/<ISSUE-KEY>
+// * Issue view: https://<YOUR-SUBDOMAIN>.atlassian.net/browse/<ISSUE-KEY>
+
+import client from '../client';
+
+const DOMAIN = '.atlassian.net';
+
+const issuesTabPattern = /^\/projects\/[A-Za-z0-9\-_]+\/issues\/[A-Za-z0-9\-_]+$/;
+const browseIssuePattern = /^\/browse\/[A-Za-z0-9\-_]+$/;
```
pmeinhardt: 🤔 These patterns might be a little to decipher when we're a little less restrictive:

```javascript
const issuesTabPattern = /^\/projects\/[^/]+\/issues\/[^/]+$/
```

pmeinhardt: I'm wondering what you think about introducing a small helper, e.g. using/based on [micro-match](https://github.com/nblackburn/micro-match) or [matchit](https://github.com/lukeed/matchit) or a similar library to support more familiar URL patterns:

```javascript
import match from 'micro-match';

const { id } = match('/projects/:project/issues/:id', path);
```

For me this feels a lot more readable plus we get the values directly instead of RegExp-test first, then extract the last segment in a separate step.

src/common/client.js
```diff
@@ -0,0 +1,16 @@
...
+const credentials = 'include';
+
+const headers = {
+  'content-type': 'application/json; charset=utf-8',
+  accept: 'application/json',
+};
+
+function client(base, options = {}) {
+  return ky.extend({
+    credentials, headers, prefixUrl: base, ...options,
```
pmeinhardt: On the way to the office, I had a more general thought… (Sorry for bringing up so much here, it's just that this is the first adapter we're moving over to use an API and to work asynchronously so naturally there's a lot of new things to think about 🤗).

Previously, all our adapters were synchronous, so the adapters and thus the UI would return results very quickly.

When we're doing async requests, these may take a moment longer which means Tickety-Tick will not be showing anything instantly. This means to things:

1. We should update the UI to display an **activity indicator** while adapters are scanning for tickets (could be a separate PR)
2. We should potentially **enforce timeouts** on adapters (can also be dealt with in a separate issue, but let's think it through)

Currently in `src/common/search.js` we simply use `Promise.all` to wait for all adapters to finish, meaning the slowest adapter determines how long we wait. Usually we only expect only one adapter to return results though. Therefore [`Promise.race`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/race) might actually be a better fit.

The simplest but not necessarily most robust approach might be to just set a default timeout for the `ky` client instances, e.g. 1000 ms?

This also leads to the question of how to deal with failed requests and rejected promises. In the simplest case, we don't handle this at all and trust all adapters to never fail and to resolve with `null` instead so they don't disturb the search. Or we could wrap all adapter promises in `search.js` and make sure they simply resolve with `null` in case they're rejected/throw an error.

src/common/adapters/jira.test.js
```diff
@@ -1,5 +1,3 @@
-import { URLSearchParams } from 'whatwg-url';
-
```
pmeinhardt: I think we also don't need to have the package as a dependency (since it's polyfilled by JSDOM).

src/common/adapters/jira.js
nickgnd: Ohh I didn't know about, I'll try it right now!

src/common/client.js
pmeinhardt: Ah… `Promise.race` doesn't quite fit by the way, because what we need is the first non-rejected, non-null result whereas `Promise.race` just resolves or rejects with the result of the promise that finishes first, no matter whether it was resolved or rejected. We can find a solution anyway 🙂

nickgnd:
> However, this means we need to update the adapter guard to check for something other than host.endsWith('.atlassian.net'). Do you have any ideas?

@pmeinhardt @andreasknoepfle can the jira host be set in the `tickety-tick` settings?

And we need also to check if the self-hosted version exposes the JSON endpoints 😨 It seems that the last release does it 😎

src/common/client.js
nickgnd: @pmeinhardt I set the timeout for the HTTP client, I guess we can address the other points in a separated PR.

pmeinhardt:
> @pmeinhardt @andreasknoepfle can the jira host be set in the `tickety-tick` settings?

I would like to avoid adding such adapter-specific configuration values as these are easily forgotten and ideally Tickety-Tick should work on any Jira instance you come to without requiring you to change the settings each time.

I think there's a few things we should assume to be given, e.g. a somewhat up-to-date Jira instance that has these APIs available. We'll likely never be able to/never want to make Tickety-Tick work with all versions of a ticket system in parallel I guess. **Let's keep it simple** — after all, we have to maintain this tool and it's not like thousands of users are depending on it at the moment. If so, we'd have to depend more on external contributions anyway.

andreasknoepfle:
> I think there's a few things we should assume to be given, e.g. a somewhat up-to-date Jira instance that has these APIs available. We'll likely never be able to/never want to make Tickety-Tick work with all versions of a ticket system in parallel I guess.

So no more tickety-tick on Coup :(

pmeinhardt:
@andreasknoepfle, we've added support for self-hosted instances. However, we'd require them to run at least Jira 8.0+. Can you provide any specifics as to which version they're running?

pmeinhardt:
@nickgnd, one thing we lose here (I think you already pointed this out) is that we do not handle selection of multiple tickets  anymore. The URLs don't seem to reflect multiple selected issues… 😢  We could still achieve this through a combination of scraping and API requests… Not sure how important this feature is compared to a - hopefully - more robust adapter using the API.

Any thoughts @klappradla, @tessi, @lwassermann?

nickgnd:
@pmeinhardt @andreasknoepfle

nickgnd:
> @andreasknoepfle, we've added support for self-hosted instances. However, we'd require them to run at least Jira 8.0+. Can you provide any specifics as to which version they're running?

@pmeinhardt Andi wrote me this morning that they are using Jira v.7.2.7, which, if I'm looking at the right documentation, it exposes REST APIs (https://docs.atlassian.com/software/jira/docs/api/REST/7.2.7/) 🤔 

What is surprising to me is that the base endpoint is different:
- in the adapter is set to: `/rest/agile/1.0`
- in the documentation that I linked above is: `/rest/api/2/`

Then, I found [this](https://docs.atlassian.com/software/jira/docs/api/REST/8.0.0/) documentation for the version 8.0 which doesn't mention the `/rest/agile/` endpoint at all, but It uses the same as the version 7.x (`/rest/api/2`)

- https://developer.atlassian.com/cloud/jira/software/rest/  (👈  the one that I looked at the beginning)
- https://docs.atlassian.com/software/jira/docs/api/REST/8.0.0/
- https://docs.atlassian.com/software/jira/docs/api/REST/7.2.7/

...And the mystery deepens. It needs a bit of investigation.

> one thing we lose here (I think you already pointed this out) is that we do not handle selection of multiple tickets anymore. The URLs don't seem to reflect multiple selected issues… 😢 We could still achieve this through a combination of scraping and API requests… Not sure how important this feature is compared to a - hopefully - more robust adapter using the API.

@pmeinhardt I personally never used this feature, so I vote to drop it. I don't like the idea of a mixed approach.

tessi:
I introduced the multi-select feature back then to easy joblog generation for project here (select all tickets you worked on that day, tickety-tick-copy them ,done. it was great times, but i also stopped using it since i'm now working on many different projects per day anyways.

so I'd be ok with dropping this feature if it makes our life easier.

pmeinhardt:
Awesome. Thanks @tessi for your feedback on the multi-select.

I think what's  missing now is:

- [ ] Feedback from @andreasknoepfle about whether this works with slightly older on-premise/self-hosted Jira instances
- [ ] Testing on Opera and Safari as @nickgnd pointed out

pmeinhardt:
> I think what's missing now is:
> 
>     * [ ] Feedback from @andreasknoepfle about whether this works with slightly older on-premise/self-hosted Jira instances
>     * [ ] …

Ping @andreasknoepfle

maltoe:
@pmeinhardt Does not seem to work in chrome with Coup's Jira :/

Screenshot 2019-03-15 at 13 02 33

Chrome: Version 72.0.3626.121 (Official Build) (64-bit)
Jira: Atlassian JIRA Project Management Software (v7.2.7#72009-sha1:68b7d86)

pmeinhardt:
Cool. Thanks a lot @maltoe. @andreasknoepfle was here for lunch and he told me that this Jira instance is hosted with a path prefix of `/jira`, which conflicts with how the adapter currently pieces together the API URLs, i.e. hostname + `/api/…`.

So we'll need to add some logic to figure out the proper base URL which includes potential path prefixes like that.

pmeinhardt:
@nickgnd, I added support for custom path prefixes on self-managed Jira instances (see last commit).
I briefly sat together with @andreasknoepfle and we confirmed it's working.

I've already rebased and from my point of view, this is good to go 🚢

PR #139, opened 2019-03-06
Additions for the new Jira adapter

pmeinhardt:
* Add support for self-hosted Jira instances (as long as they use the REST API)
* Tests that API requests are made against the correct endpoints/host
* Small refactoring

### DIFF
--- a/src/common/adapters/jira.js
+++ b/src/common/adapters/jira.js
@@ -13,43 +13,36 @@ import match from 'micro-match';
 
 import client from '../client';
 
-const DOMAIN = '.atlassian.net';
-
-const issueTabMatch = '/projects/:project/issues/:id';
-const browseIssueMatch = '/browse/:id';
-
-function selectedIssue({ href, pathname }) {
-  let id = null;
-
+function getSelectedIssueId({ href, pathname }) {
   const { searchParams: params } = new URL(href);
-  id = params.get('selectedIssue');
-  if (id) return id;
 
-  ({ id } = { ...match(issueTabMatch, pathname), ...match(browseIssueMatch, pathname) });
+  if (params.has('selectedIssue')) return params.get('selectedIssue');
 
-  return id;
+  return ['/projects/:project/issues/:id', '/browse/:id']
+    .map(pattern => match(pattern, pathname).id)
+    .find(Boolean);
 }
 
 function extractTicketInfo(response) {
-  const { key: id, fields } = response;
-  const { issuetype, summary: title } = fields;
+  const { key: id, fields: { issuetype, summary: title } } = response;
   const type = issuetype.name.toLowerCase();
-
   return { id, title, type };
 }
 
 async function scan(loc) {
   const { host } = loc;
-  if (!host.endsWith(DOMAIN)) return null;
 
-  const issueKey = selectedIssue(loc);
-  if (!issueKey) return null;
+  if (!host.endsWith('.atlassian.net')) return null;
+
+  const id = getSelectedIssueId(loc);
+
+  if (!id) return null;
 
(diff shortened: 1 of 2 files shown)


src/common/adapters/jira.js
```diff
@@ -13,43 +13,40 @@ import match from 'micro-match';
...
-const issueTabMatch = '/projects/:project/issues/:id';
-const browseIssueMatch = '/browse/:id';
-
-function selectedIssue({ href, pathname }) {
-  let id = null;
+function isJiraPage(loc, doc) {
+  if (loc.host.endsWith('.atlassian.net')) return true;
+  if (doc.body.id === 'jira') return true;
+  return false;
+}
```
nickgnd: 👍 Maybe we can update the documentation of the adapter then

src/common/adapters/jira.js
```diff
@@ -13,43 +13,40 @@ import match from 'micro-match';
...
-  id = params.get('selectedIssue');
-  if (id) return id;
 
-  ({ id } = { ...match(issueTabMatch, pathname), ...match(browseIssueMatch, pathname) });
+  if (params.has('selectedIssue')) return params.get('selectedIssue');
 
-  return id;
+  return ['/projects/:project/issues/:id', '/browse/:id']
+    .map(pattern => match(pattern, pathname).id)
+    .find(Boolean);
```
nickgnd: 👍

src/common/adapters/jira.test.js
```diff
@@ -1,26 +1,30 @@
...
+const ticket = {
+  id: response.key,
+  title: response.fields.summary,
+  type: response.fields.issuetype.name.toLowerCase(),
+};
 
 describe('jira adapter', () => {
-  function loc(host, pathname = '', params = null) {
-    const searchParams = new URLSearchParams(params || {});
+  function loc(host, pathname = '', params = {}) {
```
nickgnd: I set the default value of `params` to `null` because otherwise in the ternary operator below it is always true, I mean:

```js
const href = params ? `https://...?params` : `https://...`;
```

I see these alternatives:
- we can remove the ternary operator at all, there will be always the `?` at the end of the host, but doesn't really matter
- we can check that its keys length is more than 0 `Object.keys(params).length > 0 ? ... : ...`
- we can rollback these changes to the previous implementation

I don't have a strong opinion about that.

src/common/adapters/jira.test.js
```diff
@@ -33,68 +37,60 @@ describe('jira adapter', () => {
...
+  const doc = dom.window.document;
+
+  const api = { get: jest.fn() };
 
   beforeEach(() => {
-    client.mockImplementation(() => ({
-      get: requestMock,
-    }));
+    api.get.mockReturnValue({ json: () => response });
+    client.mockReturnValue(api);
```
nickgnd: 👍

src/common/adapters/jira.test.js
pmeinhardt: I see now. I missed that yesterday evening. Thanks, I'll revert it.

src/common/adapters/jira.js
pmeinhardt: Not sure what you mean? This just checks whether we're on a Jira page or not. The issue key/id retrieval still works the same as before.

src/common/adapters/jira.test.js
pmeinhardt: Done 👍

src/common/adapters/jira.js
nickgnd: Nevermind, I meant the maybe we can specify that the adapter can work also with a self-hosted instance. But I think is not really needed.

nickgnd:
:shipit: 👏

PR #145, opened 2019-03-14
Update dependencies

pmeinhardt:
Just updated all dependencies to their latest versions.

The extension still works as before (checked in Firefox & Chrome):

preview


.circleci/config.yml
```diff
@@ -10,12 +10,12 @@ jobs:
...
       - checkout
       - restore_cache:
-          key: dependency-cache-{{ checksum "package.json" }}
+          key: dependency-cache-{{ checksum "yarn.lock" }}-{{ arch }}
       - run:
           name: Install dependencies
           command: yarn install --frozen-lockfile
       - save_cache:
-          key: dependency-cache-{{ checksum "package.json" }}
+          key: dependency-cache-{{ checksum "yarn.lock" }}-{{ arch }}
```
nickgnd: 💡 

Thanks for the explanation in the commit message

package.json
```diff
@@ -32,6 +31,7 @@
     "babel-jest": "^23.6.0",
     "babel-loader": "8.0.5",
     "chrome-launcher": "0.10.5",
+    "clean-webpack-plugin": "^2.0.1",
```
nickgnd: 👍

jest.config.js
```diff
@@ -4,7 +4,7 @@
 module.exports = {
   coverageDirectory: 'coverage',
   collectCoverageFrom: ['<rootDir>/src/**/*.{js,jsx}'],
-  setupTestFrameworkScriptFile: '<rootDir>/test/setup.js',
+  setupFilesAfterEnv: ['<rootDir>/test/setup.js'],
```
nickgnd: Is it a new jest property? 🤔

pmeinhardt: Yep. The previous one, which accepted only a single path, was deprecated in favor of this new options that allows you to pass an array of paths.

For reference, see the note on:
https://jestjs.io/docs/en/configuration.html#setupfilesafterenv-array

pmeinhardt:
Thanks a lot for your review @nickgnd ❤️

Issue #143, opened 2019-03-07
Add a loading indicator


pmeinhardt:
With the new async adapters which may perform API requests etc. it makes sense to show an **activity indicator** while the adapters are scanning for tickets.

See https://github.com/bitcrowd/tickety-tick/pull/137#discussion_r262505287

pmeinhardt:
Closed via #146

PR #146, opened 2019-03-15
[#143] Add a loading indicator

pmeinhardt:
Add a super simple, pure CSS spinner that is shown while tickets are being loaded.

**Preview**

preview

_Note_: I added an extra timeout for demo purposes. In reality users should never see it this long.

Also, the slight stutter in the preview is due to the GIF framerate.

### DIFF
--- a/src/common/popup/popup.html
+++ b/src/common/popup/popup.html
@@ -5,6 +5,12 @@
     <title>Tickety-Tick</title>
   </head>
   <body>
-    <div id="popup-root" class="popup"></div>
+    <div id="popup-root" class="popup">
+      <div class="spinner">
+        <div class="bounce bounce-1"></div>
+        <div class="bounce bounce-2"></div>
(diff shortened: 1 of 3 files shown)


pmeinhardt:
There are certainly more inspiring spinners, even pure CSS ones. But since it'll hardly ever be on screen for a notable time, I didn't spend much time on it.

For reference: http://tobiasahlin.com/spinkit/

pmeinhardt:
> So, the spinner is always shown, doesn't matter if the adapter doesn't perform any HTTP requests or if still waiting for a response. I think that it is completely fine as the first implementation, then we can think about how to improve it if needed

Yes, exactly. It's shown immediately when you open the popup. In most cases you won't notice it, because the current adapters `scan` promises resolve right away. With the API adapter implementations though you might see it for whatever time we use as a timeout. ⏲

Issue #144, opened 2019-03-07
Handle adapter errors


pmeinhardt:
With the new API-based adapters, we'll likely have cases where requests time out or fail. For such cases, we should check that Tickety-Tick displays a helpful error message:

https://github.com/bitcrowd/tickety-tick/pull/137#discussion_r262505287

PR #150, opened 2019-03-17
[#144] Handle adapter errors 

pmeinhardt:
Okay. So after a long wait, here's a first take on notifying users when something goes wrong while scanning for tickets. I have an idea for even better error reporting which I'll try out – but that'll be a separate PR.

preview

### DIFF
--- a/src/common/popup/components/no-tickets.jsx
+++ b/src/common/popup/components/no-tickets.jsx
@@ -1,12 +1,21 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 
 import ExternalLink from './external-link';
 
-const href = 'https://github.com/bitcrowd/tickety-tick/issues';
+import ErrorShape from '../utils/error-shape';
 
-function NoTickets() {
+function IssueLink() {
   return (
-    <div className="text-center">
+    <ExternalLink href="https://github.com/bitcrowd/tickety-tick/issues">
+      Report an issue
+    </ExternalLink>
+  );
+}
+
+function Hint() {
+  return (
+    <>
       <h5 className="mt-3 mb-3">No tickets found on this page.</h5>
       <h6>
         Did you select or open any tickets?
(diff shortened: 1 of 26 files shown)


pmeinhardt:
Once this is in, I think we're good to go for a new release 🎉 
`3.0` since we now have API-based adapters?

src/common/popup/components/no-tickets.jsx
```diff
@@ -20,12 +29,47 @@ function NoTickets() {
...
+
+function Report({ errors }) {
+  return (
+    <>
+      <h5 className="mt-3 mb-3">Houston, we have a problem.</h5>
+      <p>
+        Tickety-Tick encountered
+        {' '}
+        <strong>
+          {errors.length === 1 ? 'an error' : `${errors.length} errors`}
```
klappradla: 👍 makes sense!

src/common/popup/components/tool.jsx
```diff
@@ -4,25 +4,26 @@ import PropTypes from 'prop-types';
...
 import NoTickets from './no-tickets';
 import Header from './header';
+
 import TicketShape from '../utils/ticket-shape';
+import ErrorShape from '../utils/error-shape';
 
-function Tool({ tickets }) {
-  const content = (tickets && tickets.length > 0)
+function Tool({ tickets, errors }) {
+  const content = tickets.length > 0
```
klappradla: Perfect! So we're only ever considering the errors if there are not tickets? 👍

src/common/search.js
```diff
@@ -5,11 +5,30 @@ import Ora from './adapters/ora';
...
+async function attempt(scan, loc, doc) {
+  try {
+    const tickets = await scan(loc, doc);
+    return { tickets };
+  } catch (error) {
+    return { error: serializable(error) };
+  }
+}
+
+function aggregate(results) {
```
klappradla: 💚 for the naming

src/common/utils/serializable-errors.test.js
```diff
@@ -0,0 +1,23 @@
+import serializable from './serializable-errors';
+
+describe('serializable-errors', () => {
+  const error = new Error('Serialized Test Error');
+
+  let representation;
+
+  function roundtrip(value) {
```
klappradla: Interesting approach 🤓

src/common/popup/components/tool.jsx
pmeinhardt: Exactly. Our idea from the last hacknight was that as long as a matching adapter finds tickets, it's okay for other adapters to reject with errors. So only if we don't receive any tickets, but we see errors, we report them.

src/common/utils/serializable-errors.test.js
pmeinhardt: Haha 😁 Yeah, I was wondering why I couldn't retrieve the error messages and stack traces in the pop-up window for a bit before it hit me… when they're passed from the content script to the pop-up window, they need to be serializable.

```javascript
JSON.stringify(new Error('Boom!')) === '{}'
```

PR #151, opened 2019-03-20
[#93] Enhance commit messages (part 2)

pmeinhardt:
Pretty-print commit messages according to recommendations:
https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

This isn't yet applied to the generated commit messages.

My idea was to add a checkbox to the options (enabled by default) to apply the recommended format to generated commit messages.

Let me know what you think 💚

### DIFF
--- a/src/common/format/pretty-print.js
+++ b/src/common/format/pretty-print.js
@@ -1,0 +1,56 @@
+import prettier from 'prettier/standalone';
+import markdown from 'prettier/parser-markdown';
+import unindent from 'strip-indent';
+
+const widths = { subject: 50, body: 72 };
+
+const config = { parser: 'markdown', plugins: [markdown] };
+
+function format(text, width) {
+  const options = { ...config, printWidth: width, proseWrap: 'always' };
+  return prettier.format(text, options);
+}
+
+function splitf(string, separator) {
+  const position = string.indexOf(separator);
+
+  if (position < 0) return [string];
+
+  const head = string.substring(0, position);
+  const tail = string.substring(position + separator.length);
+
+  return [head, tail];
+}
+
+function capitalizef(string) {
+  return string.replace(/[a-zA-Z]/, w => w.toUpperCase());
+}
+
+function gitsubject(string) {
+  const subject = capitalizef(string.trim());
+
+  if (subject.length > widths.subject) {
+    const [start, rest] = splitf(format(subject, widths.subject - 1), '\n');
+    return [`${start}…`, format(`…${rest}`, widths.body)].join('\n\n');
+  }
+
+  return subject;
+}
+
+function gitbody(string) {
+  const body = unindent(string.replace(/^(\s*\n)*|\s*$/, ''));
+  return format(body, widths.body);
+}
+
+function maybe(value, fn) {
+  if (typeof value !== 'string') return value;
+  return fn(value);
+}
+
+function print(input) {
+  const [line0, rest] = splitf(input.trim(), '\n');
+  const parts = [maybe(line0, gitsubject), maybe(rest, gitbody)];
+  return parts.filter(Boolean).join('\n\n');
+}
+
+export default print;
--- a/src/common/format/pretty-print.js
+++ b/src/common/format/pretty-print.js
@@ -11,7 +11,7 @@ function format(text, width) {
   return prettier.format(text, options);
 }
 
-function splitf(string, separator) {
+function split(string, separator) {
   const position = string.indexOf(separator);
 
   if (position < 0) return [string];
@@ -22,15 +22,15 @@ function splitf(string, separator) {
   return [head, tail];
 }
 
-function capitalizef(string) {
+function capitalize(string) {
   return string.replace(/[a-zA-Z]/, w => w.toUpperCase());
 }
 
 function gitsubject(string) {
-  const subject = capitalizef(string.trim());
+  const subject = capitalize(string.trim());
 
   if (subject.length > widths.subject) {
-    const [start, rest] = splitf(format(subject, widths.subject - 1), '\n');
+    const [start, rest] = split(format(subject, widths.subject - 1), '\n');
     return [`${start}…`, format(`…${rest}`, widths.body)].join('\n\n');
   }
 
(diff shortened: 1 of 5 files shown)


src/common/format/pretty-print.js
```diff
@@ -0,0 +1,56 @@
...
+  const tail = string.substring(position + separator.length);
+
+  return [head, tail];
+}
+
+function capitalizef(string) {
+  return string.replace(/[a-zA-Z]/, w => w.toUpperCase());
+}
+
+function gitsubject(string) {
```
klappradla: Also potentially interesting here: remove a dot at the end of the subject line _(the subject should not be a sentence)_

src/common/format/pretty-print.js
```diff
@@ -0,0 +1,56 @@
...
+  const subject = capitalizef(string.trim());
+
+  if (subject.length > widths.subject) {
```
klappradla: Interesting idea 👍 . While probably way too radical for e.g. a linter / autoformatter for commit messages, I feel like this is rather inspiring / encouraging here. Like it!

src/common/format/pretty-print.js
```diff
@@ -0,0 +1,56 @@
...
+}
+
+function maybe(value, fn) {
+  if (typeof value !== 'string') return value;
+  return fn(value);
+}
+
+function print(input) {
+  const [line0, rest] = splitf(input.trim(), '\n');
+  const parts = [maybe(line0, gitsubject), maybe(rest, gitbody)];
```
klappradla: Love the naming here 💚

src/common/format/pretty-print.js
```diff
@@ -0,0 +1,56 @@
...
+  return parts.filter(Boolean).join('\n\n');
```
klappradla: Never seen this syntax 🤓 

My assumption: it filters out NULLs / undefined elements?

klappradla:
PS: not sure if you're already working on top of that PR / have a feature branch, etc. So I don't dare to merge myself potentially causing you branch chaos @pmeinhardt ✌️

src/common/format/pretty-print.js
pmeinhardt: Yes, your intuition is right on point! It's a pattern you come across in many JS packages these days.

src/common/format/pretty-print.js
pmeinhardt: I am not sure, but I think there might be linters/linter plugins in projects like danger etc.

I didn't find a formatter though which is why I wrote this thing and – like you – I was suprised it didn't take a whole lot of code thanks to projects like prettier 🙌

pmeinhardt: …we could release it as a separate `pretty-print-git-message` package though (the name is completely a placeholder) if you think that's useful?

src/common/format/pretty-print.js
pmeinhardt: Ah, I saw that in a post somewhere but didn't bother to include it here, assuming that most ticket descriptions aren't using a dot anyway 🤔 Do you think we should add that?

src/common/format/pretty-print.js
klappradla: 🎉 nice idea! Not sure about the extra maintenance effort though. Of couse, if everythink works fine, we rarely have to touch it. But if some leftpad party or anything else happens, it's one more thing to take care of...

src/common/format/pretty-print.js
klappradla: Sure, but for the sake of consistency I feel like it still makes sense to add it. If we already opened the bucket of "Git Subject" then this is just a tiny extra on top.

pmeinhardt:
Thanks for your awesome feedback @klappradla ☀️ I'm seriously really glad you like this and like you I was stoked to see how efficient something like this could be implemented 😀

src/common/format/pretty-print.js
pmeinhardt: Cool! I'll add it when I have time 🙏 Thank you!

pmeinhardt:
TODO:

- [ ] strip dot in subject

pmeinhardt:
Note: Just rebased to keep in sync with `master`.

src/common/format/pretty-print.js
pmeinhardt: Mmh, I thought I'd quickly add stripping of end-of-sentence punctuation marks for the Git subject lines… 🤔 

However, looking at it more closely it looks like a can of worms. To illustrate, here's a bunch of new tests I started out with:

```javascript
  it('strips trailing end-of-a-sentence punctuation marks', () => {
    expect(print('Strip full stops.')).toBe('Strip full stops');
    expect(print('Strip exclamation marks!')).toBe('Strip exclamation marks');
    expect(print('Strip question marks?')).toBe('Strip question marks');
    expect(print('Keep ellipsis...')).toBe('Keep ellipsis...');
    expect(print('Keep ellipsis…')).toBe('Keep ellipsis…');
  });
```

Also, I've had one existing test case that started failing and to me it's hard to handle such cases properly:

```javascript
  ● pretty-print › wraps overlong subject lines

    expect(value).toMatchSnapshot()

    Received value does not match stored snapshot "pretty-print wraps overlong subject lines 1".

    - Snapshot
    + Received

      "Wrap commit subject lines with more than 50…

      …characters and insert one blank line before the remaining subject text.
    - Wrap the remaining subject text to 72 characters.
    + Wrap the remaining subject text to 72 characters
      "

      17 |   it('wraps overlong subject lines', () => {
      18 |     const input = 'Wrap commit subject lines with more than 50 characters and insert one blank line before the remaining subject text. Wrap the remaining subject text to 72 characters.';
    > 19 |     expect(print(input)).toMatchSnapshot();
         |                          ^
      20 |   });
      21 |
      22 |   it('wraps overlong body lines', () => {

      at Object.toMatchSnapshot (src/common/format/pretty-print.test.js:19:26)
```

In this case, stripping the full stop just makes it look super-odd.

I'm currently leaning more towards leaving punctuation marks alone. What do you think @klappradla?

pmeinhardt: @klappradla, this probably drowned between all the mails you received in your absence. ☝️
Maybe you can take another look? 🙏

klappradla: Sorry. I still have ~200 Github notification emails and wasn't brave enough to really tackle them yet...

I agree with what you wrote: let's not open this can of worms! If it's nothing that's done "easily" the naive way, then starting without it and looking at the "real world" usage makes more sense I'd say ✅

pmeinhardt: Cool, thanks a lot @klappradla and I hope you'll make quick progress working through that impressive backlog of GitHub notifications 🤞😀

PR #156, opened 2019-05-18
[#144] Handle adapter errors (part 2)

pmeinhardt:
Here's an attempt at improving error reporting for adapter errors.

1. It uses the source-maps we ship with the bundled extension to create a clear stacktrace.
2. It allows users to easily copy the error message + stacktrace.

Ideally this makes it easier for users (including us) to provide more detailed issue descriptions.

preview

To record this demo ☝️ I just inserted a `throw new Error('Boom!');` into the GitHub adapter.

### DIFF
--- a/src/common/popup/components/error-details.jsx
+++ b/src/common/popup/components/error-details.jsx
@@ -1,0 +1,65 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import StackTrace from 'stacktrace-js';
+
+import CopyButton from './copy-button';
+
+import ErrorShape from '../utils/error-shape';
+
+function framefmt(stackframe) {
+  const name = stackframe.getFunctionName() || '<anonymous>';
+  const file = stackframe.getFileName() || '-';
+  const line = stackframe.getLineNumber() || '-';
+  const column = stackframe.getColumnNumber() || '-';
+  return `    at ${name} (${file}:${line}:${column})`;
+}
+
+async function trace(error) {
+  const stackframes = await StackTrace.fromError(error); // parse and enhance with source maps
+  return [error.message, ...stackframes.map(framefmt)].join('\n');
+}
+
+class ErrorDetails extends Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      traces: null,
+    };
+  }
+
+  componentDidMount() {
+    const { errors } = this.props;
+
+    Promise.all(errors.map(trace))
+      .then(traces => this.setState({ traces }))
+      .catch(err => console.error(err));
+  }
+
+  render() {
+    const { traces } = this.state;
+
+    if (!traces) return null;
+
+    const preamble = `Tickety-Tick revision: ${COMMITHASH}`;
+    const logs = traces.map(tr => ['```', tr, '```'].join('\n'));
+
+    const info = [preamble, ...logs].join('\n\n');
+
+    return (
+      <CopyButton
+        className="btn btn-primary"
+        title="Copy error details to help identify the issue"
+        value={info}
+      >
+        Copy error details
+      </CopyButton>
+    );
+  }
+}
+
+ErrorDetails.propTypes = {
+  errors: PropTypes.arrayOf(ErrorShape.isRequired).isRequired,
+};
+
+export default ErrorDetails;
--- a/package.json
+++ b/package.json
@@ -76,6 +76,7 @@
     "react-router-dom": "5.0.0",
     "react-textarea-autosize": "^7.1.0",
     "speakingurl": "14.0.1",
+    "stacktrace-js": "^2.0.0",
     "strip-indent": "^2.0.0",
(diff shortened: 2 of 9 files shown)


src/common/popup/components/error-details.jsx
```diff
@@ -0,0 +1,50 @@
...
+
+    const stack = stackframes.map(sf => sf.toString()).join('\n');
+    const info = [error.message, stack].join('\n\n');
+
+    return (
+      <div className="p-2 text-left">
+        <p className="font-weight-bold">{error.message}</p>
+        <pre className="overflow-auto">
+          <code>{stack}</code>
+        </pre>
```
andreasknoepfle: does it maybe make sense to use a readonly textarea instead, so it is scrollable by default? 
At least this approach seems quite common :)

src/common/popup/components/error-details.jsx
```diff
@@ -0,0 +1,50 @@
...
+      stackframes: null,
+    };
+  }
+
+  componentDidMount() {
+    const { error } = this.props;
+
+    StackTrace.fromError(error)
+      .then(stackframes => this.setState({ stackframes }))
+      .catch(err => console.error(err));
```
andreasknoepfle: 👍 for outputting it on the console still :)

src/common/popup/components/error-details.jsx
```diff
@@ -0,0 +1,50 @@
...
+    const stack = stackframes.map(sf => sf.toString()).join('\n');
+    const info = [error.message, stack].join('\n\n');
+
+    return (
+      <div className="p-2 text-left">
+        <p className="font-weight-bold">{error.message}</p>
+        <pre className="overflow-auto">
+          <code>{stack}</code>
+        </pre>
+        <CopyButton value={info}>Copy error details</CopyButton>
```
andreasknoepfle: It is so fun that we could reuse this one :D

src/common/popup/components/error-details.jsx
```diff
@@ -0,0 +1,50 @@
...
+      </div>
+    );
+  }
+}
+
+ErrorDetails.propTypes = {
+  error: ErrorShape.isRequired,
+};
+
+export default ErrorDetails;
```
andreasknoepfle: should we add a snapshot test for this?

src/common/popup/components/error-details.jsx
pmeinhardt: Not sure what benefit a textarea would give us here, esp. because we'd then need to style the textarea to use a monospaced code font etc.

src/common/popup/components/error-details.jsx
pmeinhardt: What do you mean?

src/common/popup/components/error-details.jsx
pmeinhardt: 👍 I'll definitely add tests. This first shot was just to get people's feedback before I follow through 😁 Also the UI needs a little more work beforehand/

src/common/popup/components/error-details.jsx
andreasknoepfle: I mean it is fun that we already have a copy button around as a default feature of tickety tick. In other projects one would have needed to build this stuff :D.  
No action required.

pmeinhardt: Haha, I see. Yep 😁

src/common/popup/components/error-details.jsx
andreasknoepfle: just thought it may be visually nicer, since the text is wierdly cut and it may not be obvious for a user that this is scrollable. But on the other hand 🤷‍♂  not to important :D

src/common/popup/components/error-details.jsx
pmeinhardt: This is only logging an error in case the stacktrace enhancement (apply source maps) fails.

pmeinhardt:
The stacktraces in Safari aren't as helpful, but hey… 🤷‍♂

At least for the other browsers it'll hopefully make issues easier to find.

---

Tickety-Tick revision: 0fa52d9927757323abbc9c6df33c314e754f6fb8

```
Boom!
    at <anonymous> (safari-extension://net.bitcrowd.tickety-tick-0000000000/996db8f5/content.js:3:6788)
    at asyncFunctionResume ([native code]:-:-)
    at <anonymous> (safari-extension://net.bitcrowd.tickety-tick-0000000000/996db8f5/content.js:3:7925)
    at asyncFunctionResume ([native code]:-:-)
    at map ([native code]:-:-)
    at <anonymous> (safari-extension://net.bitcrowd.tickety-tick-0000000000/996db8f5/content.js:3:7872)
    at asyncFunctionResume ([native code]:-:-)
    at <anonymous> (safari-extension://net.bitcrowd.tickety-tick-0000000000/996db8f5/content.js:3:9955)
```

PR #157, opened 2019-05-19
Show version information in pop-up

pmeinhardt:
Display the Git commit hash on the pop-up info screen and link to the
repo tree for that commit of the extension.

preview

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -5,6 +5,9 @@
     "browser": true,
     "jest": true
   },
+  "globals": {
+    "COMMITHASH": "readonly"
+  },
   "plugins": [
     "jest"
   ],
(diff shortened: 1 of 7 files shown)


pmeinhardt:
Checked out [git-revision-webpack-plugin](https://github.com/pirelenito/git-revision-webpack-plugin) which seems really useful. I've always wanted something like this so you know exactly which version of the software you're looking at, e.g. for issue reports.  👀 

This could be something worth integrating in many of our client projects as well 🤔

pmeinhardt:
In particular, this could be something worth integrating on our current gig as well @lwassermann 😉

pmeinhardt:
Additional note: This could also be nice to integrate with Sentry error reporting etc. 💡

Issue #165, opened 2019-05-31
Unable to copy default value in settings


maltoe:
Hello friends,

just a minor annoyance I just noticed, not even sure this is TT's fault or just chrome code:

When I try to change a setting in the Chrome extension settings, say the formatter's `branch` setting, I can't copy the existing value as it is just a placeholder in the input.

![ticketytick_cant_copy](https://user-images.githubusercontent.com/96114/58692536-1822b300-838f-11e9-8cf5-9bd02ffaa583.gif)

❤️ 
malte

pmeinhardt:
Ah, indeed. I totally see why that's slightly frustrating @maltoe 💚 Great screencast 😄 
The default is just shown as a standard browser [input placeholder](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#Labels_and_placeholders) which is not selectable.

I guess the alternative would be to set the input value itself to the default value on blur if the field is empty. Would that be the behavior you'd prefer @maltoe?

Also pinging @lwassermann here who might have more input?

maltoe:
@pmeinhardt Or just prefill the field with the default, and don't persist it if it hasn't been changed?

andreasknoepfle:
We could also prefill it and have a reset button or something like this, to go to be able to go to the default again.

pmeinhardt:
Fixed by #170.

PR #170, opened 2019-06-06
[#165] Unable to copy default value in settings 

pmeinhardt:
preview

### DIFF
--- a/src/web-extension/options/components/template-input.jsx
+++ b/src/web-extension/options/components/template-input.jsx
@@ -14,6 +14,8 @@ TemplateInputElement.propTypes = {
   multiline: PropTypes.bool.isRequired,
 };
 
+const noop = () => {};
+
 function TemplateInput(props) {
   const {
     id,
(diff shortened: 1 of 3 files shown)


pmeinhardt:
@maltoe and @andreasknoepfle thank you both for your feedback 👍 
I've quickly put together a hacky version of a solution.

Take a look or - better - take it for a spin and let me know what you think 👋

pmeinhardt:
Okay, so here's an updated version with tests. The behavior seems intuitive to me at least 😁

maltoe:
@pmeinhardt Tested, works for me, and intuitive enough. Possibly increase the font size of the default just a tiny bit? Feel free to ingore.

pmeinhardt:
I've bumped up the base font size for the options page a little bit. For some reason there's a huge gap in the rendered font sizes in Firefox and Chrome, so bumping it up even more makes it look super-awkward in Chrome. I tried to hit some middleground there that still seems somewhat reasonable.

preview

Thanks for your feedback @maltoe 👍

pmeinhardt:
This'll go into the `3.2` release: https://github.com/bitcrowd/tickety-tick/milestone/7?closed=1
@andreasknoepfle, maybe you can find a quiet moment today to publish? ✌️

PR #174, opened 2019-06-16
Fix "node binary" build warnings

pmeinhardt:
There was a warning when executing `package.json` scripts that used `npm run` to execute other scripts. This is fixed by using `run-s` from [`npm-run-all`](https://github.com/mysticatea/npm-run-all) instead.

```
tickety-tick ❯ yarn bundle:chrome
yarn run v1.16.0
$ cross-env BUNDLE=true npm run build:chrome
npm WARN lifecycle The node binary used for scripts is /var/folders/95/l40glm311hl4w31f8c2n3g6h0000gn/T/yarn--1560718780495-0.6512468251304615/node but npm is using /usr/local/Cellar/node/12.4.0/bin/node itself. Use the `--scripts-prepend-node-path` option to include the path for the node binary npm was executed with.

> tickety-tick@3.1.0 build:chrome /Users/pmeinhardt/Code/bitcrowd/tickety-tick
> cross-env VARIANT=chrome webpack --mode=production --config webpack.webext.babel.js
…
```


package.json
```diff
@@ -10,13 +10,13 @@
...
-    "watch:safari": "npm run build:safari -- --watch",
+    "watch:chrome": "run-s build:chrome -- --watch",
+    "watch:firefox": "run-s build:firefox -- --watch",
+    "watch:safari": "run-s build:safari -- --watch",
     "open:chrome": "./script/open-in-chrome ./dist/chrome https://github.com/bitcrowd/tickety-tick",
     "open:firefox": "web-ext run --source-dir ./dist/firefox --pref startup.homepage_welcome_url=https://github.com/bitcrowd/tickety-tick",
-    "bundle:chrome": "cross-env BUNDLE=true npm run build:chrome",
-    "bundle:firefox": "cross-env BUNDLE=true npm run build:firefox",
+    "bundle:chrome": "cross-env BUNDLE=true run-s build:chrome",
+    "bundle:firefox": "cross-env BUNDLE=true run-s build:firefox",
```
andreasknoepfle: 🤔 isn't the `run-s` from `npm-run-all` usually used to run mutliple npm commands at once or something? 

I am also wondering, since we seem to use `yarn` if this should actually call `yarn` here instead of `npm` anyway (correct me if I am wrong, since I wander on thin ice here :D)

pmeinhardt: Yes, using `yarn` inside the scripts should also work. `run-s` automatically determines whether you're using `npm` or `yarn` to run a script from the `package.json` so I figured I'd just use that and we don't have to bother, even when someone else decides to use `npm run bundle:chrome` or similar 🤷‍♂

Issue #178, opened 2019-06-20
Automate release process


andreasknoepfle:
We could make it easier to release tickety tick by automating it a bit better. 
- [x] Add a task for zipping the safari extension
- [x] Add a script to draft a new release and upload the bundles to github (maybe for running in CI on new tags)

Optional: 
- [x] Use the commit history/changelog to create release descriptions

klappradla:
Some work-in-progress notes on that:

Following up on a discussion I had with @pmeinhardt on https://github.com/bitcrowd/tickety-tick/issues/212, I quickly tried out [`np`](https://github.com/sindresorhus/np) introduced in #184 to automatically create Github releases. It works perfectly fine for that, but as expected does not support adding downloadable release artifacts 😞  _(people already wanted to use it for [uploading extensions](https://github.com/sindresorhus/np/issues/92))_

What works pretty neat is this Github Action: https://github.com/actions/upload-release-asset. The [example workflow](https://github.com/actions/create-release#example-workflow---create-a-release) essentially combines two actions (one for creating the release and one for adding a downloadable artifact to the new release).

We'll probably need to combine a few tools. Some more are:
- [chrome-webstore-upload-cli](https://github.com/DrewML/chrome-webstore-upload-cli)
- [web-ext-sign](https://extensionworkshop.com/documentation/develop/web-ext-command-reference/#web-ext-sign)

Here is an [example workflow](https://github.com/npmhub/npmhub/blob/955beed238d85fe41a60c04d01dc99dfe85bd6c6/.github/workflows/deploy-tag.yml) from another extension project which does the two uploads.

klappradla:
Done as a semi automated, Capistrano-like flow 🐘

metaqer:
hello

metaqer:
hello

PR #184, opened 2019-08-01
[#178] Automate release process (2)

pmeinhardt:
Check out [`np`](https://github.com/sindresorhus/np) for releasing Tickety-Tick ✌️ 

⚠️  Note: The webpack `GitRevisionPlugin` is not outputting the correct `COMMITHASH` and `VERSION` yet, because the `npm/yarn version` script runs just before commiting.


klappradla:
I think it makes sense to pick this up again for setting up the automation flow I roughly outlined in https://github.com/bitcrowd/tickety-tick/issues/178#issuecomment-624559141 @pmeinhardt. If we want to trigger build magic on CI based on pushing tags, utilizing `yarn release` and `np` could be a starting point.

Given we currently don't actually support Safari any more, I would probably already get rid of the extra parts for it.

pmeinhardt:
Thanks a ton @klappradla for looking into release automation. 💚

I'd actually be fine with dropping Safari support. It simplifies a lot of things (webpack config, repo structure, preferences… reduces extra abstractions in app initialization/React context…).

Also, it seems like a new, compatible Safari version would essentially be a completely different code base with Apple deciding to go "native app extensions" rather sticking to the WebExtension standards. https://github.com/bitcrowd/tickety-tick/issues/210

PR #189, opened 2019-09-11
Use keyboard input element in README

klappradla:
Updates the section on keyboard shortcuts to output the keys using the
awesome [Keyboard Input HTML element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/kbd).

⬇️ looks like this (the rich diff) ⬇️ 
Screenshot 2019-09-11 at 11 49 08

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,9 +47,9 @@ Tickety-Tick is available for every major browser:
 
 ## Keyboard Shortcuts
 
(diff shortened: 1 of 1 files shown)


pmeinhardt:
Love it ❤️ Thanks @klappradla!

PR #198, opened 2020-01-17
[#195] Support backlog & board view of JIRA next gen software projects

### LINKED ISSUE #195: Support backlog & board view of JIRA next gen software projects
JIRA has a new type of project which we do not yet support fully:

https://confluence.atlassian.com/jirasoftwarecloud/working-with-next-gen-software-projects-945104895.html

An example is e.g. our own bchp board: 
https://bitcrowd.atlassian.net/jira/software/projects/BCHP/boards/7

Both views currently make tickety-tick error out: 

Screenshot 2019-11-28 at 18 47 10

juffel:
Add next gen project url structure to regex of known url suffixes. These URLs look like this:

https://bitcrowd.atlassian.net/jira/software/projects/BCHP/boards/7?selectedIssue=BCHP-149

This fixes #195 

TODO:
- [x] add a test for this kind of path in `jira.test.js`

### DIFF
--- a/src/common/adapters/jira.js
+++ b/src/common/adapters/jira.js
@@ -19,11 +19,7 @@ function isJiraPage(loc, doc) {
   return false;
 }
 
-const pathSuffixes = new RegExp(
-  "/(browse/[^/]+|projects/[^/]+/issues/[^/]+|secure/RapidBoard.jspa)|(jira/software/projects/[^/]+/boards/[^/]+/)$",
-  "g"
-);
-
+const pathSuffixes = new RegExp("/(browse/[^/]+|projects/[^/]+/issues/[^/]+|secure/RapidBoard.jspa|jira/software/projects/[^/]+/boards/[^/]+/)$", "g");
 function getPathPrefix(loc) {
   return loc.pathname.replace(pathSuffixes, "");
 }
@@ -35,15 +31,15 @@ function getSelectedIssueId(loc, prefix = "") {
 
   const path = loc.pathname.substr(prefix.length); // strip path prefix
 
-  return ["/projects/:project/issues/:id", "/browse/:id"]
-    .map(pattern => match(pattern, path).id)
-    .find(Boolean);
+  return (["/projects/:project/issues/:id", "/browse/:id"]
+    .map((pattern) => match(pattern, path).id)
+    .find(Boolean));
 }
 
 function extractTicketInfo(response) {
   const {
     key: id,
-    fields: { issuetype, summary: title }
+    fields: { issuetype, summary: title },
   } = response;
   const type = issuetype.name.toLowerCase();
   return { id, title, type };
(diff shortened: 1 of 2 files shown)


src/common/adapters/jira.js
```diff
@@ -19,7 +19,7 @@ function isJiraPage(loc, doc) {
   return false;
 }
 
-const pathSuffixes = new RegExp('/(browse/[^/]+|projects/[^/]+/issues/[^/]+|secure/RapidBoard.jspa)$', 'g');
+const pathSuffixes = new RegExp('/(browse/[^/]+|projects/[^/]+/issues/[^/]+|secure/RapidBoard.jspa)|(jira/software/projects/[^/]+/boards/[^/]+/)$', 'g');
```
pmeinhardt: Is the `jira/software/…` part new or optional? All the other example paths above don't seem to have that (even though the regexes would still match. So I'm wondering whether the regex should require the prefix. What happens if you remove `jira/software` from the URL in your browser? Does it still work?

qclin: `jira/software/projects` is the path prefix when you browsing from the panel navigation jumping from Boards/Roadmap/Backlog views

pmeinhardt: I see. Thanks @qclin. 💚 

The `RegExp` can be simplified slightly, to be:

```suggestion
const pathSuffixes = new RegExp('/(browse/[^/]+|projects/[^/]+/issues/[^/]+|secure/RapidBoard.jspa|jira/software/projects/[^/]+/boards/[^/]+/)$', 'g');
```

and we need a test for this.

src/common/adapters/jira.test.js
```diff
@@ -76,6 +76,15 @@ describe('jira adapter', () => {
     expect(result).toEqual([ticket]);
   });
 
+
+  it('extracts tickets from new generation software projects', async () => {
+    const result = await scan(loc('my-subdomain.atlassian.net', `/jira/software/projects/TT/boards/8/backlog?selectedIssue=${key}`), doc);
```
pmeinhardt: The `loc` test helper isn't really written in the best way (sorry 🙄), but for consistency, we should do as on line 62:

```suggestion
    const result = await scan(loc('my-subdomain.atlassian.net', '/jira/software/projects/TT/boards/8/backlog', `?selectedIssue=${key}`), doc);
```

src/common/adapters/jira.test.js
```diff
@@ -76,6 +76,15 @@ describe('jira adapter', () => {
     expect(result).toEqual([ticket]);
   });
 
+
+  it('extracts tickets from new generation software projects', async () => {
+    const result = await scan(loc('my-subdomain.atlassian.net', '/jira/software/projects/TT/boards/8/backlog', `?selectedIssue=${key}`), doc);
+
```
pmeinhardt:
```suggestion
```

PR #200, opened 2020-02-25
Upgrade postcss-preset-env version

klappradla:
This updates the version of the `postcss-preset-env` package we use as a development dependency.

Background: running the tests showed the following deprecation warning:

> Browserslist: caniuse-lite is outdated. Please run next command `yarn upgrade`

The `caniuse-lite` is a dependency of the `postcss-preset-env` package - therefore I upgraded that.


pmeinhardt:
Thank you 💚

Issue #199, opened 2020-02-25
Drop support for Pivotal Tracker


klappradla:
As reported some time ago in #89, our current adapter for [Pivotal Tracker](https://www.pivotaltracker.com/) is broken 😿 

As there aren't any active users of this tracker in our community, we at least for now unfortunately have to drop the support for Pivotal 🕳

PR #201, opened 2020-02-25
[#199] Drop support for Pivotal Tracker

klappradla:
This removes Pivotal Tracker from the list of supported ticket systems as described in #199:
- remove the adapter and related tests
- adjust the docs to no longer list pivotal tracker

While updating the list of supported trackers, I also added a small note to encourage people to implement support for Pivotal by themselves if they need it 🤷‍♂ Let's see.

Up next would be: draft a release, release it and adjust the descriptions in the respective places (Chrome / Firefox).

It would also be neat if someone with Github superpowers could remove the `pivotal-tracker` tag from the description of this repo 😉


src/common/search.js
```diff
@@ -17,10 +16,13 @@ async function attempt(scan, loc, doc) {
 }
 
 function aggregate(results) {
-  return results.reduce(({ tickets, errors }, result) => ({
-    errors: errors.concat(result.error ? [result.error] : []),
-    tickets: tickets.concat(result.tickets || []),
-  }), { tickets: [], errors: [] });
+  return results.reduce(
```
klappradla: Note: this is only a side-effect of the linter having fun after line 32 got shorter 😉

src/common/adapters/pivotal.js
```diff
@@ -1,55 +0,0 @@
-import {
-  $all,
-  $classed,
```
juffel: Looks like we could now even throw out some functions from `src/common/adapters/helpers.js` like `classed` and `closest` 🗑

README.md
```diff
@@ -33,9 +33,10 @@ Currently, we support:
 - [GitLab](https://gitlab.com/)
 - [Jira](https://www.atlassian.com/software/jira)
 - [Ora](https://ora.pm/)
-- [Pivotal Tracker](https://www.pivotaltracker.com/)
 - [Trello](https://trello.com/)
 
+Your ticket system is missing? Go ahead and implement an adapter for it! We are always happy about contributions and here to support you 👋.
```
pmeinhardt: 😍

juffel: 👍

src/common/adapters/pivotal.js
klappradla: Neat! I checked which ones were not used and removed them ✅

pmeinhardt:
Awesome @klappradla. Good to be merged from my point of view 🙂

klappradla:
It was quite some time since the last release @pmeinhardt 🤔 

I'll release a new version, ok?

pmeinhardt:
I think the "summary" button might not quite work as before (since we have the full commit messages now, not just the title). I don't think a lot of people are using that anyway though and we can address that later. 🤔

PR #215, opened 2020-03-09
Update authors to be all of us

tessi:
Since bodo os not working at bitcrowd anymore, the mail address was out of date in package.json. This changes it to our tickety-tick address we use to register with the addon stores. Also it does not refer to a specific person anymore but all bitcrowd devs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "3.2.0",
   "name": "tickety-tick",
   "description": "A browser extension that helps you to create commit messages and branch names from story trackers.",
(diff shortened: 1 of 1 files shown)


klappradla:
Not sure where and how the name is displayed, but alternatives in the same direction would be:
- dev team at bitcrowd
- bitcrowd

🤷‍♂

tessi:
Hmm i am ok with (almost) any alternative phrases. How do we decide for one?

pmeinhardt:
Might be nice to mention Bodo as the initial/original creator of the idea.

pmeinhardt:
I applied @klappradla's suggestion. ✌️

PR #216, opened 2020-03-12
[#195] Support backlog & board view of JIRA next…

pmeinhardt:
…gen software projects.

Closes #195.
Might also fix #214.

The Jira path RegExp was not matching and thus the requested API URL was not constructed correctly, i.e. we got:

```
https://bitcrowd.atlassian.net/jira/software/projects/BCHP/boards/7/rest/api/latest/issue/BCHP-211
```

instead of

```
https://bitcrowd.atlassian.net/rest/api/latest/issue/BCHP-211
```

**TODO:** Add a test case

PR #217, opened 2020-04-09
Fix jira URL prefixes to work with new jira boards

### LINKED ISSUE #195: Support backlog & board view of JIRA next gen software projects

tessi:
The JIRA board URLS changed for new JIRA boards. They now contain
optional trailing slashes and also allow further nested routes.
Examples:

* `/jira/software/projects/ABCD/boards/7`
* `/jira/software/projects/ABCD/boards/7/backlog`

Where before, only this URL was recognized:

* `/jira/software/projects/BCHP/boards/7/`

Closes #195.

### DIFF
--- a/src/common/adapters/jira.js
+++ b/src/common/adapters/jira.js
@@ -34,9 +34,9 @@ function getSelectedIssueId(loc, prefix = '') {
 
   const path = loc.pathname.substr(prefix.length); // strip path prefix
 
-  return (['/projects/:project/issues/:id', '/browse/:id']
+  return ['/projects/:project/issues/:id', '/browse/:id']
     .map((pattern) => match(pattern, path).id)
-    .find(Boolean));
+    .find(Boolean);
 }
 
 function extractTicketInfo(response) {
--- a/src/common/adapters/jira.js
+++ b/src/common/adapters/jira.js
@@ -19,7 +19,10 @@ function isJiraPage(loc, doc) {
   return false;
 }
 
-const pathSuffixes = new RegExp('/(browse/[^/]+|projects/[^/]+/issues/[^/]+|secure/RapidBoard.jspa|jira/software/projects/[^/]+/boards/[^/]+/)$', 'g');
+const pathSuffixes = new RegExp(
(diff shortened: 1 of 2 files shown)


src/common/adapters/jira.js
```diff
@@ -19,7 +19,10 @@ function isJiraPage(loc, doc) {
   return false;
 }
 
-const pathSuffixes = new RegExp('/(browse/[^/]+|projects/[^/]+/issues/[^/]+|secure/RapidBoard.jspa|jira/software/projects/[^/]+/boards/[^/]+/)$', 'g');
+const pathSuffixes = new RegExp(
+  '/(browse/[^/]+|projects/[^/]+/issues/[^/]+|secure/RapidBoard.jspa|jira/software/projects/[^/]+/boards/.*)$',
```
tessi, line 23: `jira/software/projects/[^/]+/boards/.*` captures all possible trailing URL parts of the board view. I opted to not be very specific here and just catch all possible routes within `/boards/` to be more stable against further JIRA changes.

pmeinhardt:
Is this a different problem from: https://github.com/bitcrowd/tickety-tick/pull/216?

tessi:
> Is this a different problem from: #216?

This is the exact issue i fixed - by not scanning (and removing) the prefix path correctly that weird URL from #216 gets built.

tessi:
@pmeinhardt true that, added tests. Actually one of the existing tests passed although the wrong jira base_url was called. I added a check against which API URL is called to the relevant tests.

src/common/adapters/jira.js
pmeinhardt: Cool. This should also cover what we had here: https://github.com/bitcrowd/tickety-tick/pull/216/files#diff-a4549fad9112b44c628079812d266381L22-R22

PR #220, opened 2020-04-12
Remove build flags with default values

pmeinhardt:
~Only apply build optimizations for creating releases or if specifically
requested. This roughly cuts build times in half for development builds.~

Remove webpack build flags that were using the defaults.


pmeinhardt:
> looks good, is this disabling code ~obfuscation~ minifaction for dev builds? Would make debugging the extension much easier

Thanks for asking about it.

I double-checked the documentation and it turns out, `'production'` is actually the default mode. I tried again and removing/re-adding the flag doesn't make a difference.

I'll check again later, to see whether setting `--mode=development` instead makes a difference.

pmeinhardt:
Here's a link to the docs for reference: https://webpack.js.org/configuration/mode/

package.json
```diff
@@ -7,16 +7,16 @@
   "license": "MIT",
   "scripts": {
     "build": "run-s -n build:chrome build:firefox build:safari",
-    "build:chrome": "cross-env VARIANT=chrome webpack --mode=production --config webpack.webext.babel.js",
-    "build:firefox": "cross-env VARIANT=firefox webpack --mode=production --config webpack.webext.babel.js",
+    "build:chrome": "cross-env VARIANT=chrome webpack --config webpack.webext.babel.js",
+    "build:firefox": "cross-env VARIANT=firefox webpack --config webpack.webext.babel.js",
     "build:safari": "webpack --mode=production --config webpack.safari.babel.js",
```
pmeinhardt, line 12: Note to self: Forgot to remove the flag for `build:safari` here. ☝️

package.json
```diff
@@ -7,16 +7,16 @@
   "license": "MIT",
   "scripts": {
     "build": "run-s -n build:chrome build:firefox build:safari",
-    "build:chrome": "cross-env VARIANT=chrome webpack --mode=production --config webpack.webext.babel.js",
-    "build:firefox": "cross-env VARIANT=firefox webpack --mode=production --config webpack.webext.babel.js",
+    "build:chrome": "cross-env VARIANT=chrome webpack --config webpack.webext.babel.js",
```
klappradla, line 10: Hm, I honestly don't understand how this makes a difference 🙈 

According to [the docs](https://webpack.js.org/configuration/mode/):
> If not set, webpack sets production as the default value for mode.

Reading this, I'd have expected the build to be 1:1 the same 🤔

pmeinhardt: Yup. As noted above ✌️ The speed-up initially measured seems to have come from somewhere else. I've dabbled with this a bit more yesterday and wasn't able to get any improvements even when using `--mode=development`. For now, I'll just update the PR to remove the flags (since they're the default anyway).

klappradla: Oh, sorry for the duplicate! Did not see your comment above 🙈

README.md
```diff
@@ -61,6 +61,12 @@ yarn checks
 yarn build
 ```
 
+To build an [un-optimized version for development or debugging](https://webpack.js.org/configuration/mode/), use:
+
+```shell
+yarn build --mode=development
```
klappradla, line 67: Another thing I was wondering: how does `yarn` even know this flag is part of the `webpack` command? Looking at the `package.json`, `build` runs two command which then both call webpack 🤯

pmeinhardt: Thanks for bringing that up. Yarn and npm both simply append extra arguments to the scripts. It's `run-s` that takes care of adding the arguments to all of the individual scripts:
https://github.com/mysticatea/npm-run-all/blob/master/docs/run-s.md

The previous version here didn't work though, because you need to add a placeholder. I updated the PR accordingly. Thank you 💚 

🙇

klappradla: 🤯 _(but the tool is super powerful)_

Issue #163, opened 2019-05-30
Please add notion.so support :)


xhr15:
It would be lovely to use tickety-tick with notion.so

pmeinhardt:
We could maybe do a lunch coding session maybe to spread a bit of knowledge on how to work on Tickety-Tick, some general JS tooling etc. 🙂

xhr15:
it turned out notion is:
1) hard to parse
2)saying that an API is on the roadmap, but there is no progress

Therefore we are dropping notion support for now

pmeinhardt:
Notion is likely back for a new project, so let's see whether we can come up with a way of supporting it. ✌️

pmeinhardt:
So, here's my findings so far:

- Looks like we cannot really tell what a "ticket" is in notion, because they keep everything super generic and don't have a notion (pun intended) of what is a "ticket".
- However, it looks like everything is basically a "page", so we could generate commit messages and branch names based on pages.

Which information do we need?

- ticket identifier
- ticket title
- (optional) type, like "feature", "bug"…
- (optional) description
- (optional) URL

How could we go about it?

1. Starting from the URL, we can derive the page ID (UUID)

    - When directly on the ticket page, the URL looks like

      `https://www.notion.so/<organization name>/<slugified title>-<page uuid, no dashes>`

      Turns out you can drop the `slugified title` and the URL still works. This would mean we can construct a shorter (optional) URL to include in the ticket information.

    - When opened in an overlay, the URL follows this pattern

      `https://www.notion.so/<organization name>/<parent uuid, no dashes>?v=<version?>&p=<page uuid, no dashes>`

    - So the "proper" page UUID can be constructed by inserting the dashes in the appropriate places, e.g.

      `96ec637de4b04a5eacf38d4d9a1b2e4b` becomes `96ec637d-e4b0-4a5e-acf3-8d4d9a1b2e4b`

2. We then send a `POST` request to `https://www.notion.so/api/v3/getRecordValues` (or, since we're on the same domain, `/api/v3/getRecordValues`)

    - The request body should contain:

      ```json
      {
        "requests": [
          { "table": "block", "id": "96ec637d-e4b0-4a5e-acf3-8d4d9a1b2e4b" }
        ]
      }
      ```

    - This will return a JSON response like the following:


      ```json
      {
        "results": [
          {
            "role": "editor",
            "value": {
              "id": "96ec637d-e4b0-4a5e-acf3-8d4d9a1b2e4b",
[...]

PR #224, opened 2020-04-15
[#163] Please add notion.so support :)

klappradla:
This adds support for the [Notion](www.notion.so) tracker as beautifully prepared by @pmeinhardt in #163 🙏 💚 

Notion does not differentiate between "stories", "tasks", etc. For them, all actionable entites are **"pages"**. Tickety-Tick therefore also support creating branch names and commit messages for pages.

See here in action👇 

notion-small

Please note: the board view is not an actionable "page" and therefore not supported by Tickety-Tick 🧑‍🏫 .

The commit messages for now only have a subject containing the rather long ID of the page and its title. For the page from the screencast this is:

**branch**

```
page/5b1d7dd7-9107-4890-b2ec-83175b8eda83-add-image-block
```

**message**

```
[#5b1d7dd7-9107-4890-b2ec-83175b8eda83] Add image…

…block
```

**command**

```
git checkout -b 'page/5b1d7dd7-9107-4890-b2ec-83175b8eda83-add-image-block' && git commit --allow-empty -m '[#5b1d7dd7-9107-4890-b2ec-83175b8eda83] Add image…

…block
'
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,8 +32,8 @@ Currently, we support:
 - [GitHub](https://github.com/)
 - [GitLab](https://gitlab.com/)
 - [Jira](https://www.atlassian.com/software/jira)
-- [Trello](https://trello.com/)
 - [Notion](https://www.notion.so/)
+- [Trello](https://trello.com/)
 
 Your ticket system is missing? Go ahead and implement an adapter for it! We are always happy about contributions and here to support you 👋.
 
--- a/README.md
+++ b/README.md
@@ -33,6 +33,7 @@ Currently, we support:
 - [GitLab](https://gitlab.com/)
 - [Jira](https://www.atlassian.com/software/jira)
 - [Trello](https://trello.com/)
+- [Notion](https://www.notion.so/)
 
 Your ticket system is missing? Go ahead and implement an adapter for it! We are always happy about contributions and here to support you 👋.
 
--- a/src/common/adapters/notion.js
+++ b/src/common/adapters/notion.js
@@ -54,9 +54,11 @@ function extractTicketInfo(result) {
   return { id, title, type };
 }
 
-function getTickets(response) {
-  const { results = [] } = response;
-  return results.map(extractTicketInfo).filter(Boolean);
+function getTickets(response, id) {
+  return (response.results || [])
+    .map(extractTicketInfo)
+    .filter(Boolean)
+    .filter((t) => t.id === id);
 }
 
 async function scan(loc) {
@@ -70,7 +72,7 @@ async function scan(loc) {
   const request = { json: { requests: [{ table: 'block', id }] } };
   const response = await api.post('api/v3/getRecordValues', request).json();
 
-  return getTickets(response);
+  return getTickets(response, id);
 }
 
 export default scan;
--- a/src/common/adapters/notion.js
+++ b/src/common/adapters/notion.js
@@ -14,16 +14,16 @@ import { match } from 'micro-match';
 import client from '../client';
 
 /**
- * Turns a page ID without dashes into a dasherized RFC 4122 UUID.
+ * Turns a slugged page ID without dashes into a dasherized RFC 4122 UUID.
  * UUID-Format: 96ec637d-e4b0-4a5e-acf3-8d4d9a1b2e4b
  */
-function uuid(string) {
+function uuid(slugId) {
   return [
-    string.substring(0, 8),
-    string.substring(8, 12),
-    string.substring(12, 16),
-    string.substring(16, 20),
-    string.substring(20),
+    slugId.substring(0, 8),
+    slugId.substring(8, 12),
+    slugId.substring(12, 16),
+    slugId.substring(16, 20),
+    slugId.substring(20),
   ].join('-');
 }
 
@@ -37,10 +37,10 @@ function getPageFromPath(path) {
 function getSelectedPageId(loc) {
   const { pathname: path, searchParams: params } = new URL(loc.href);
   const isPageModal = params.has('p');
-  const id = isPageModal ? params.get('p') : getPageFromPath(path);
+  const slugId = isPageModal ? params.get('p') : getPageFromPath(path);
 
-  if (!id) return null;
-  return uuid(id);
+  if (!slugId) return null;
+  return uuid(slugId);
 }
 
 function extractTicketInfo(result) {
--- a/src/common/adapters/notion.js
+++ b/src/common/adapters/notion.js
@@ -1,0 +1,76 @@
+/**
+ * notion.so adapter
+ *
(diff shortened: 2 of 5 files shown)


src/common/adapters/notion.js
```diff
@@ -0,0 +1,76 @@
...
+  const { id, type } = value;
+  if (type !== 'page') return null;
+
+  const title = value.properties.title[0][0];
+  return { id, title, type };
+}
+
+function getTickets(response) {
+  const { results = [] } = response;
+  return results.map(extractTicketInfo).filter(Boolean);
```
klappradla, line 59: This would essentially also return multiple tickets if the response would yield multiple items in the `results` object. As `scan` is supposed to return a collection, I found this convenient. But it may also be a source for hidden problems in the future 🤔

src/common/adapters/notion.js
```diff
@@ -0,0 +1,76 @@
+/**
+ * notion.so adapter
+ *
+ * The adapter extracts the UUID of a selected notion.so task ("page") from the
+ * page URL and uses the notion.so API to retrieve the corresponding task
+ * information.
+ *
+ * Note: this adapter uses notion.so's internal v3 API as there is no public
+ * API yet.
+ */
```
pmeinhardt, line 10: Neat documentation. 👌

I noticed I usually just use `//` for commenting, even when the comment consists of multiple lines. 🙈 I found this convenient, when your (previously) one-line comment becomes longer and suddenly spans multiple lines.

However, I also checked the Airbnb [styleguide](https://github.com/airbnb/javascript#comments) which we're using as a basis and they say to use `/** … */`for multi-line comments.

Do you have any preference/opinion? Should we update those other multi-line comments (in a separate PR)? 🤔

src/common/adapters/notion.js
pmeinhardt: 🤔 Indeed. My initial thought when writing the proposal was to `results.filter` and match by the UUID that we've constructed before mapping. Something along the lines of:

```js
function getTickets(response, id) {
  return (response.results || []).filter(r => r.id === id).map(extractTicketInfo);
}
```

README.md
```diff
@@ -33,6 +33,7 @@ Currently, we support:
 - [GitLab](https://gitlab.com/)
 - [Jira](https://www.atlassian.com/software/jira)
 - [Trello](https://trello.com/)
+- [Notion](https://www.notion.so/)
```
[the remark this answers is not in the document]
klappradla: There is no "fair" order. But let's go with alphabetical, so it's at least not our fault that it's not fair.

src/common/adapters/notion.test.js
```diff
@@ -0,0 +1,111 @@
+import client from '../client';
+import loc from './__helpers__/location';
+import scan from './notion';
+
+jest.mock('../client', () => jest.fn());
```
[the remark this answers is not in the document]
klappradla: Cool. Will take care of it. I feel like I'm slowly getting the hang of Jest's mocking 🤞

src/common/adapters/notion.test.js
```diff
@@ -0,0 +1,111 @@
...
+    const request = { requests: [{ table: 'block', id }] };
+    expect(api.post).toHaveBeenCalledWith('api/v3/getRecordValues', {
+      json: request,
+    });
+    expect(result).toEqual([]);
+  });
+
+  it('returns an emtpy array when the page does not exist', async () => {
+    const res = { results: [{ role: 'editor' }] };
+    api.post.mockReturnValueOnce({ json: () => res });
```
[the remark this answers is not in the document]
klappradla: Yes, at least for this scenario. There may be cases of "actual" 404s on the API but I did not discover them yet.

I tried out a few different error scenarios with the type of request we're doing by replaying modified requests with `fetch` in the console. Our payload roughly looks like this:
```js
{
  requests: [
    {
      table: 'block',
      id: 'some-uuid'
    }
  ]
}
```
Regardless of what I do to the `id` field value, changing it to an invalid UUID _(not confirm to RFC 4122)_, changing it to a dash or whitespace, I'll always get a `200` success response. An empty one though.

Changing the value of the `table` field, e.g. to `admins` will return a `400` response though. The payload could look somewhat like this:
```js

{
  requests: [
    { table: 'admins', id: 'some-uuid' }
  ]
}
```
The response is then an object with an `errorId`, `message`, etc. I initially also had this as a test scenario, but then realized `ky` is treating status codes outside of the 2xx range as an error anyways 🤷. So I guess it already does everything we need?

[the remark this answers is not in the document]
pmeinhardt: I see. Thanks a lot for the detailed explanation. 🙇 I was just wondering whether the test should use `mockRejectedValue` exactly because `ky` throws for non-`2xx` responses. But if we get a success, it's perfect as it is. 💚

src/common/adapters/notion.js
klappradla: I find this closer to how documentation looks like in Elixir and felt like it's a better fit for these documentation snippets on top of files. I discovered it in the wordpress and [google](https://google.github.io/styleguide/jsguide.html#jsdoc-top-file-level-comments) styleguides as well.

No strong feelings, but if you think it's visually cleaner, I can also update the rest of the class comments. Would also be a good opportunity to check whether they're still up to date 😉

src/common/adapters/notion.js
klappradla: This means we're tightening the contract for `scan` to in the success case always return an array with only a single item. Sounds good and is probably a better safety-net for discovering changes to their API or general problems with our adapter.

Will update accordingly 🆗

klappradla: Implemented your suggestion _(thanks for the tip to skip the object deconstruction 🎉 )_ but kept the filter for null values.

Background: the ID is nested in the `value` object of the `result`. So filtering before extracting the ticket information would already do most of the work that's done when actually extracting the information. What do you think?

pmeinhardt: Looks neat 👌

src/common/adapters/notion.test.js
```diff
@@ -85,6 +85,22 @@ describe('notion adapter', () => {
     expect(result).toEqual([]);
   });
 
+  it('returns an empty one if the page id does not match the requested one', async () => {
```
pmeinhardt, line 88: Should this say:

```suggestion
  it('returns an empty array if the page id does not match the requested one', async () => {
```

klappradla: Whoopsy 🙈

src/common/adapters/notion.test.js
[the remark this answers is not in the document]
klappradla: Taking care of this in https://github.com/bitcrowd/tickety-tick/issues/225

Issue #176, opened 2019-06-20
Maybe add a border glow to the icon in the menu bar?


bitboxer:
When I am using dark mode, i barely see the tickety tick icon in the menu bar. Maybe adding a glow border would help it standing out?

pmeinhardt:
Pretty much all browsers use a rather flat UI style without any gradients, shadows or glows.
I'd therefore prefer not to add a glow.

I wonder if there's a way for either providing "light" and "dark" icon versions (e.g. via the extension manifest) or an icon mask that is colored by the browser (something that Apple does for iOS app navbars…)?

🤔

pmeinhardt:
Mmh. There doesn't seem to be a way of providing a different light and dark icons:

- https://developer.chrome.com/extensions/manifest/icons
- https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/icons

Not sure if it's possible to set the icon on runtime from JavaScript and to query the browser/OS UI mode?

bitboxer:
But what about just modifying the current logo with a little glow to make it work on both dark and bright background?

This is how it looks like on my dark firefox right now:

<img width="70" alt="Screenshot 2019-09-04 at 10 36 55" src="https://user-images.githubusercontent.com/56195/64239127-f3932580-ceff-11e9-97d4-fb0feb30346d.png">

pmeinhardt:
> Pretty much all browsers use a rather flat UI style without any gradients, shadows or glows.
> I'd therefore prefer not to add a glow.

Rather than a glow, a flat light-grey circle in the background might look better. Similar to what it already looks like when you hover the icon. But adding an extra circle or other outline around it also means the logo itself will be even smaller.

![screenshot: hover](https://user-images.githubusercontent.com/706519/64240504-824c7600-cef1-11e9-8624-6dd36f2109d3.png)

If it really bothers you, why not try out a couple of variants and open a PR? 🙂 
I think the difficulty will be to find a version that works for both light and dark themes.

Personally, I also use "dark mode" but since the button stays in the same place, I don't have trouble clicking on it. Most of the time though, I just use the `ctrl+t` shortcut to open the extension popup.

Cheers 👋

PR #227, opened 2020-04-16
[#176] Support dark themes in Firefox

klappradla:
This addresses a point @bitboxer mentioned in #176: the dark icon may be "too dark" when using a dark theme or macOS' dark mode.

This is a suggestion on how we could support this _(at least for Firefox)_ 🦊 using the `theme_icons` property in [`browser_action`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/browser_action).  
By configuring this, we can show a light menu icon when users use the dark Firefox theme (or macOS "dark mode").

I also - this is only an experiment ⚠️ - changed the colors to match the ones recommended in Firefox' styleguide on [iconography](https://design.firefox.com/photon/visuals/iconography.html) 🖌️ 

The two colors are:
- dark gray: #f9f9facc
- light gray: #0c0c0dcc

This would look like this:

**dark** 🏴 
Screenshot 2020-04-16 at 18 28 30

**light** 🏳️ 
Screenshot 2020-04-16 at 18 28 39

The icon changes automatically without adding additional scripts:
firefox

**NOTE**
It would also be possible to support this behavior in Chrome. I could only reproduce this via JS and having additional events in the `background.js`... I personally find the price too high for just the color of the icon for now.

I could - if people are interested - play around with the Chrome approach in a separate PR though. Happy to do that 🤓


script/build-icons
```diff
@@ -1,11 +1,19 @@
...
+export_png_icon() {
+  src=$1 dest=$2
+  inkscape -z -o "$dest" -D -w "$s" "$src" # export
+  convert "$dest" -background none -gravity center -extent "${s}x${s}" "$dest" # pad
+  pngcrush -brute -ow "$dest" # reduce
+}
 
 for s in 16 32 48 64 128; do
-  dest="${dir}/icon-${s}.png"
-  inkscape -z -e $dest -D -w $s $src # export
```
klappradla, line 8: FYI: the `-e` flag is apparently no longer supported (at least Inkscape errors for me when I try this...) 💥 

I could not find the proper documentation, but I think™ that `-o` should be the equivalent for current versions:
```
File export:
  -o, --export-file=EXPORT-FILENAME          Output file name (file type is guessed from extension)
  --export-overwrite                         Overwrite input file
  --export-type=[...]                        File type(s) to export: [svg,png,ps,eps,pdf,tex,emf,wmf,xaml]
```

klappradla:
PS: the adaption of the `build-icons` script was only blindly throwing things in with only half my brain on... We can probably revert most of it 😉

script/build-icons
```diff
@@ -1,11 +1,19 @@
 #!/bin/sh
 
-dir=$(cd $(dirname $0) && cd "../src/common/icons" && pwd)
-src="${dir}/icon.svg"
+dir="$(cd "$(dirname "$0")" && cd "../src/common/icons" && pwd)"
```
pmeinhardt, line 3: What did my past self do here. 🤦😁 

The following should do the same in a less indirect way:

```suggestion
dir="$(realpath "$(dirname "$0")../src/common/icons")"
```

script/build-icons
```diff
@@ -1,11 +1,19 @@
 #!/bin/sh
```
pmeinhardt, line 1: I guess we should probably set some sane defaults:

```suggestion
#!/usr/bin/env bash

set -o errexit
set -o pipefail
set -o nounset

```

— https://kvz.io/bash-best-practices.html

script/build-icons
```diff
@@ -1,11 +1,19 @@
...
+dir="$(cd "$(dirname "$0")" && cd "../src/common/icons" && pwd)"
+dark_src="${dir}/icon.svg"
+light_src="${dir}/icon-light.svg"
+
+export_png_icon() {
+  src=$1 dest=$2
+  inkscape -z -o "$dest" -D -w "$s" "$src" # export
+  convert "$dest" -background none -gravity center -extent "${s}x${s}" "$dest" # pad
+  pngcrush -brute -ow "$dest" # reduce
+}
```
pmeinhardt, line 12: We should probably pass the target size as a third argument. The variable `s` is only defined below and using it here introduces a very indirect dependency.

```suggestion
function export_png_icon() {
  local src=$1
  local dest=$2
  local size=$3

  inkscape -z -o "$dest" -D -w "$size" "$src" # export
  convert "$dest" -background none -gravity center -extent "${size}x${size}" "$dest" # pad
  pngcrush -brute -ow "$dest" # reduce
}
```

script/build-icons
```diff
@@ -1,11 +1,19 @@
...
 
 for s in 16 32 48 64 128; do
-  dest="${dir}/icon-${s}.png"
-  inkscape -z -e $dest -D -w $s $src # export
-  convert $dest -background none -gravity center -extent "${s}x${s}" $dest # pad
-  pngcrush -brute -ow $dest # reduce
+  dark_dest="${dir}/icon-${s}.png"
+  light_dest="${dir}/icon-light-${s}.png"
+  export_png_icon "$dark_src" "$dark_dest"
+  export_png_icon "$light_src" "$light_dest"
```
pmeinhardt, line 18: As commented above, I think we should explicitly pass the size:

```suggestion
  export_png_icon "$dark_src" "$dark_dest" "$s"
  export_png_icon "$light_src" "$light_dest" "$s"
```

src/web-extension/manifest.json
```diff
@@ -33,7 +33,14 @@
...
     "default_title": "Git Branch/Message",
     "default_icon": "icon-128.png",
-    "default_popup": "popup.html"
+    "default_popup": "popup.html",
+    "theme_icons": [
+      {
+        "light": "icon-light-128.png",
+        "dark": "icon-128.png",
+        "size": 128
+      }
```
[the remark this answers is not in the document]
klappradla: In theory yes, but I found it quite hard in practice, so I for not just left it as it was 😉. I read the guide on [choosing icon sizes](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/browser_action#Choosing_icon_sizes) and it states:

> The browser action's icon may need to be displayed in different sizes in different contexts:
> * The icon is displayed by default in the browser toolbar, but the user can move it into the browser's menu panel (the panel that opens when the user clicks the "hamburger" icon). The icon in the toolbar is smaller than the icon in the menu panel.
> * On a high-density display like a Retina screen, icons needs to be twice as big.

☝️ this makes sense, but there's also this addition:

> If the browser can't find an icon of the right size in a given situation, it will pick the best match and scale it.

In my case it was always using the 16px one 😹 (looks hilarious - I first thought there wasn't an icon at all).

For the `default_icon` key, one could also use an SVG icon and the browser would scale it automatically to its needs. Unfortunately, the `theme_icons` don't support this though.

But I guess I'll have another look and see what's wrong with our configuration here 👀

src/common/icons/icon.svg
```diff
@@ -1 +1 @@
...
\ No newline at end of file
+<svg enable-background="new 0 0 100 100" viewBox="0 0 100 125" xmlns="http://www.w3.org/2000/svg"><path d="m76.844 50.149c0-10.038 8.156-18.178 18.156-18.204v-11.51c0-4.003-3.095-7.435-7.098-7.435h-75.581c-.55 0-1.085.176-1.601.291-.092 3.005-2.532 5.479-5.539 5.565-.106.49-.181 1.055-.181 1.579v5.65c4 0 5.689 2.546 5.689 5.689 0 3.141-1.689 5.689-5.689 5.689v7.168c4 0 5.689 2.546 5.689 5.689 0 3.141-1.689 5.689-5.689 5.689v7.168c4 0 5.689 2.546 5.689 5.689 0 3.141-1.689 5.689-5.689 5.689v5.342c0 .632.1 1.246.254 1.832 2.895.118 5.24 2.307 5.446 5.18.519.118 1.067.091 1.621.091h75.581c4.003 0 7.098-3.101 7.098-7.104v-11.543c-10-.027-18.156-8.166-18.156-18.204zm-17.233-4.999-15.396 16.534c-.921.921-2.129 1.382-3.337 1.382-1.19 0-2.379-.448-3.295-1.342-.014-.014-.028-.026-.042-.04l-6.355-6.355c-1.843-1.843-1.843-4.831 0-6.674s4.831-1.843 6.674 0l3.018 3.018 13.197-12.059c1.843-1.843 4.831-2.981 6.674-1.138s.705 4.831-1.138 6.674z" fill="#0c0c0d" fill-opacity=".8"/></svg>
```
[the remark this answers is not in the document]
klappradla: No clue, but I'll give it a shot.

Quick background on what I did to end up with these SVGs: I edited the existing SVG icon in Inkscape to change the color and then ran the result through [imageoptim](https://imageoptim.com/mac) (💚).

src/web-extension/manifest.json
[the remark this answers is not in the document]
pmeinhardt: 🤔 At least as per [MDN](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/browser_action)  we should provide at least 16 and 32 px versions. I just thought that since it's an array anyway, we could just feed all of them to the browser and let it choose on its own. Or did you mean that when you do that it doesn't actually choose a good size?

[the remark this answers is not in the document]
klappradla: Yes, the latter. When I do it, it chooses the 16px version.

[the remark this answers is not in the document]
pmeinhardt: 🤦 Oh, alright. Then it's probably best to keep it like you did, with the largest version. 👍 

Thanks for the explanation. 💚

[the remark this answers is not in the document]
klappradla: Did some more reading and I think ™ I got it right now. See d56cd4df0099e5f64eb4f2204403fe4d194e4018 (details in commit message).

The icon is now sized correctly everywhere (also in Chrome - no color change there though).

src/common/icons/icon.svg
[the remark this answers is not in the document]
klappradla: As expected: does not make any difference 👍 

I removed the attribute and - just to make sure everything is alright - re-created the icons in 2cfe8639c914607965b06b1a4379c5d0853819c4 ✅

script/build-icons
klappradla: Was missing a `/` but works like charm ✨ (did not know about `realpath`)

script/build-icons
```diff
@@ -1,11 +1,28 @@
...
```
pmeinhardt, line 27: The size doesn't seem to be passed as a third argument yet? 👀

script/build-icons
```diff
@@ -1,11 +1,28 @@
 #!/bin/sh
 
-dir=$(cd $(dirname $0) && cd "../src/common/icons" && pwd)
-src="${dir}/icon.svg"
+# script/build-icons: Build optimized png icons in all required sizes and
+#                     colors from the svg icons.
+#
+#                     Running this script requires `inkscape` and `pngcrush`
+#                     executables in your PATH.
```
pmeinhardt, line 7: The indentation seems a bit excessive 😁 

```suggestion
# Build optimized PNGs in all required sizes and colors from the SVG icons.
#
# Requires `inkscape` and `pngcrush` executables in your PATH.
```

script/build-icons
pmeinhardt: Oops. Maybe I was too early and you were still working no it? Sorry 😬

script/build-icons
klappradla: Yes! Setting these defaults is definitely 💯 

We don't necessarily have to move from `sh` to `bash` though in my opinion 🤷 . We're not really doing anything too fancy here.

pmeinhardt: > We don't necessarily have to move from `sh` to `bash` though in my opinion 🤷 . We're not really doing anything too fancy here.

Totally cool with me. Just wasn't 100% sure - from the top of my head - whether `sh` supports these options. ✌️

klappradla: I was actually surprised as well 🤓. But according to `shellcheck` it supports all but the one executing on failures within pipes (which is fine as we don't use it).

script/build-icons
klappradla: Very very good catch 👏 

Added it ✅ 

I skipped `function` and `local` though as I kept using `sh` and these super advanced features aren't available there 😉 (they are also not super critical here).

script/build-icons
klappradla: Was in a super long meeting, sorry. Should have put a `WIP` label... 🙈

script/build-icons
klappradla: 🤷 I just went with how they format scripts in [scripts-to-rule-them-all](https://github.com/github/scripts-to-rule-them-all)

script/build-icons
klappradla: Done ✅

klappradla:
Done now @pmeinhardt 🧑‍🏭 Had lots of interruptions, sorry 🙇 

Million thanks for the review by the way! Really neat points, love the outcome of this 🖤

PR #235, opened 2020-05-01
[#233] New helper to limit string length 

### LINKED ISSUE #233: new helper to limit string length
I would love to have a helper to limit string lenghts because the branch names are getting waaaay to long sometimes. So long that they break a build system I have to use for a customer right now :trollface: .

pmeinhardt:
I was intrigued @bitboxer and had a moment this morning to throw something together.

It allows for (optionally) parameterized filters in templates, e.g.:

```
Title is "{value | truncate(50)}"
```

I've also changed how we handle errors in template expressions, hoping to make it easier to figure out what's going wrong in case of typos and erroneous templates. For instance:

- The above template will now return `Title is "!!(no helper named "truncate")"` if a `truncate` helper doesn't exist
- For a template `…{value | truncate(crash)}` it will return `…!!(invalid parameters provided to "truncate": crash)`

From what I can see, there's at least two more to-dos:

- [x] We need to slightly update the "helper" documentation on the Tickety-Tick options page to inform users about available parameters.
- [x] Update template code to not break on incomplete templates `{value | }`

Please take it for a spin when you have time.

Happy about feedback. 🙂 

Closes #233.

preview

### DIFF
--- a/src/common/format/helpers.js
+++ b/src/common/format/helpers.js
@@ -1,8 +1,12 @@
 import { createSlug } from 'speakingurl';
 
-export const lowercase = (s) => s.toLowerCase();
-export const shellquote = (s) =>
+export const lowercase = () => (s) => s.toLowerCase();
+
+export const shellquote = () => (s) =>
   typeof s === 'string' ? `'${s.replace(/'/g, "'\\''")}'` : "''";
-export const slugify = createSlug({ separator: '-' });
-export const trim = (s) => s.replace(/^\s+|\s+$/g, '');
-export const uppercase = (s) => s.toUpperCase();
+
+export const slugify = () => createSlug({ separator: '-' });
+
+export const trim = () => (s) => s.trim();
+
+export const uppercase = () => (s) => s.toUpperCase();
--- a/src/common/format/helpers.js
+++ b/src/common/format/helpers.js
@@ -8,10 +8,12 @@ export const shellquote = () => (s) =>
 export const slugify = (separator = '-') => createSlug({ separator });
 
 export const substring = (...args) => (s) => s.substring(...args);
+substring.description = 'substring(start-index[, end-index])';
 
 export const trim = () => (s) => s.trim();
 
 export const truncate = (limit) => (s) =>
   s.length > limit ? `${s.substring(0, limit - 1)}…` : s;
+truncate.description = 'truncate(max-length)';
 
 export const uppercase = () => (s) => s.toUpperCase();
--- a/src/common/format/helpers.js
+++ b/src/common/format/helpers.js
@@ -5,8 +5,13 @@ export const lowercase = () => (s) => s.toLowerCase();
 export const shellquote = () => (s) =>
   typeof s === 'string' ? `'${s.replace(/'/g, "'\\''")}'` : "''";
 
-export const slugify = () => createSlug({ separator: '-' });
+export const slugify = (separator = '-') => createSlug({ separator });
+
+export const substring = (...args) => (s) => s.substring(...args);
 
 export const trim = () => (s) => s.trim();
+
+export const truncate = (limit) => (s) =>
+  s.length > limit ? `${s.substring(0, limit - 1)}…` : s;
 
 export const uppercase = () => (s) => s.toUpperCase();
--- a/src/common/format/template.js
+++ b/src/common/format/template.js
@@ -1,6 +1,53 @@
-const trim = (s) => s.replace(/^\s+|\s+$/g, '');
-const identity = (x) => x;
+const trim = (s) => s.trim();
 
+function safe(fn) {
+  return function wrapped(...args) {
+    try {
+      return fn(...args);
+    } catch (err) {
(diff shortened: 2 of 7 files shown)


src/common/format/template.js
```diff
@@ -1,6 +1,57 @@
...
+  }
+}
+
+// Compile a template string into a render function.
+//
+// Example strings:
+//
+//   'a = {v}'
+//   'b = {v | lowercase}'
+//   'c = {v | lowercase | substring(0, 3)}'
```
pmeinhardt, line 53: These examples can probably be improved a bit.

src/common/format/helpers.js
```diff
@@ -1,8 +1,17 @@
 import { createSlug } from 'speakingurl';
 
-export const lowercase = (s) => s.toLowerCase();
-export const shellquote = (s) =>
+export const lowercase = () => (s) => s.toLowerCase();
```
klappradla, line 3: Is this what smart people refer to as "currying"? Compiling functions from functions?

src/common/format/helpers.js
```diff
@@ -1,8 +1,17 @@
...
   typeof s === 'string' ? `'${s.replace(/'/g, "'\\''")}'` : "''";
-export const slugify = createSlug({ separator: '-' });
-export const trim = (s) => s.replace(/^\s+|\s+$/g, '');
-export const uppercase = (s) => s.toUpperCase();
+
+export const slugify = (separator = '-') => createSlug({ separator });
+
+export const substring = (...args) => (s) => s.substring(...args);
+
+export const trim = () => (s) => s.trim();
```
klappradla, line 12: 🥇

src/common/format/template.js
```diff
@@ -1,6 +1,53 @@
-const trim = (s) => s.replace(/^\s+|\s+$/g, '');
-const identity = (x) => x;
+const trim = (s) => s.trim();
 
+function safe(fn) {
+  return function wrapped(...args) {
```
klappradla, line 4: Just that I get things right: `wrapped` is a _"new"_ name for the thing you are returning here?

src/common/format/template.test.js
```diff
@@ -8,49 +8,81 @@ describe('template', () => {
...
-  it('handles incomplete template expressions', () => {
+  it('handles incomplete template expressions (no closing brace)', () => {
     const render = compile('{', {});
     const output = render({});
     expect(output).toBe('{');
   });
+
+  it('handles incomplete template expressions (incomplete filter pipeline)', () => {
+    const trim = () => (s) => s.trim();
+    const render = compile('{a | trim |}', { trim });
```
klappradla, line 84: Wow. Took me close to minutes to spot the trailing pipe 😹

src/common/format/template.js
pmeinhardt: Yeah, with "the thing" being a new function that runs the original function `fn` with the arguments it receives, but catches errors and turns them into strings. This makes sure that when someone is using the helper(-creators) incorrectly, the options page etc. don't break and you get a somewhat helpful output instead.

pmeinhardt: I actually wrote this without a name for the returned function first, but our linting rules (sensibly) require us to provide a name.

src/common/format/helpers.js
pmeinhardt: Not strictly, but you've got the right idea of how this works. 👍 

It's not _currying_ in the strict sense, because if you look at its definition, it's "[a translation] of a function that takes multiple arguments into evaluating a sequence of functions, each with a single argument". — https://en.wikipedia.org/wiki/Currying

Our new helper "creators" here can take more than a single argument in the first step (see `substring = (...args) => …`).

So really what we've done is turn the helper functions, and changed them from `String -> String` to `[Any] -> String -> String` (roughly) to allow configuration of helpers in the first evaluation step.

pmeinhardt:
> Plus, I agree this could use some more help on the preferences page. Even though the error feedback is already an awesome step in the right direction. Do you want to do it in this PR or tackle it as a separate one?

Maybe we can think about what the help should look like here first. At the moment, the options page lists only the helper names. This was sufficient before, when helpers did not take any arguments.

Now, we probably need at least a hint for what arguments a helper accepts. E.g. I could imagine the following for our new helpers:

- `substring(start-index[, end-index])`
- `truncate(max-length)`

We could achieve this by assigning an optional `description` property to the helpers and rendering that on the options page if defined. E.g.

```js
export const substring = (...args) => (s) => s.substring(...args);
substring.description = 'substring(start-index[, end-index])';
```

klappradla:
Sorry it took me so long to get back to this @pmeinhardt 🙈 (blog business...)

I think the format you suggested here fits perfect and I would suggest to go that route:
```
substring(start-index[, end-index])
truncate(max-length)
```

Should we have this as a separate PR and just merge this as it is?

pmeinhardt:
I've quickly added it here @klappradla (see the last commit). Looks like this

preview

Issue #212, opened 2020-02-27
Install from file not working for Firefox


klappradla:
I think™ this may actually be related to what I reported earlier in #106 🙈 : it's currently not possible to install the Firefox build as a "permanent" add-on "from file" as described in [the README](https://github.com/bitcrowd/tickety-tick#firefox).

One cannot select the `manifest.json` file for this. And trying to use the "bundle" zip file build for the submitting it to the Firefox add-ons collection also yields an error:
<img width="392" alt="Screenshot 2020-02-28 at 00 06 37" src="https://user-images.githubusercontent.com/3491815/75495248-5b618f00-59be-11ea-9c7c-ab2c998bc568.png">

It would be super helpful for the developers here ab bitcrowd to be able to run a "beta" version locally.

We should try to update our `webext` config to produce a build we can install permanently in my opinion.

pmeinhardt:
For some more pointers: https://support.mozilla.org/en-US/kb/add-on-signing-in-firefox

pmeinhardt:
It very much sounds like it is no longer possible to install self-built, unsigned extensions in "Release" and "Beta" Firefox.

https://wiki.mozilla.org/Add-ons/Extension_Signing

> What are my options if I want to install unsigned extensions in Firefox?
>
> The Nightly and Developer Edition versions of Firefox have a preference to disable signature enforcement. There are also be special unbranded versions of Release and Beta that have this preference, so that add-on developers can work on their add-ons without having to sign every build. To disable signature checks, you will need to set the `xpinstall.signatures.required` preference to "false".
> - type `about:config` into the URL bar in Firefox
> - in the Search box type `xpinstall.signatures.required`
> - double-click the preference, or right-click and selected "Toggle", to set it to `false`.

It sounds like we have the option to publish "beta" releases of Tickety-Tick:

> What about Beta versions on AMO? Will they be reviewed and signed?
>
> Yes. Beta versions will be treated like non-AMO add-ons. They will be automatically validated and signed if they pass validation.

Or maybe there's some way of self-signing them (requires access to Mozilla API):
https://extensionworkshop.com/documentation/develop/web-ext-command-reference/#web-ext-sign

This is somewhat annoying 🤔

tessi:
Alternatively, we could locally use a dev eition of firefox 🤔

klappradla:
Thanks for digging into this @pmeinhardt 🙏 (I simply wanted to report it yesterday - did not look into the details yet).

[...]

PR #238, opened 2020-05-06
[#212] Add 'sign' script and update Firefox docs

klappradla:
This addresses #212: Firefox by default does no longer allow to permanently install unsigned extensions.

Therefore, I added a `yarn sign` script for self-signing "unlisted" beta version of the extension via `web-ext` (we will most likely run this from CI in the future). According to the docs and experience on the internet, the `sign` command is not yet ready for _"listed"_ extensions. It may be possible to to both though: submit the same version as "listed" and "unlisted". We probably want to do this if we _(what I would suggest)_ also want to give people the option install the extension from a file.

I also updated the docs on that topic and will update the instructions in 1password in a minute.

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -9,5 +9,5 @@ Offering a (beta-) version for download requires self-signing the add-on through
 Find the API credentials in 1Password and run:
 
 ```shell
-yarn build:firefox && yarn sign 
+WEB_EXT_API_KEY=xxx WEB_EXT_API_SECRET=xxx yarn build:firefox && yarn sign
 ```
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -6,8 +6,8 @@ Tips & tricks for maintainers of Tickety-Tick.
 
 Offering a (beta-) version for download requires self-signing the add-on through [addons.mozilla.org](https://extensionworkshop.com/documentation/publish/#distribute-your-signed-extension). This generates a `.xpi` for installing the extension in Firefox.
 
-Find the API credentials in 1password and run:
+Find the API credentials in 1Password and run:
 
-```
+```shell
 yarn build:firefox && yarn sign 
 ```
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -9,5 +9,5 @@ Offering a (beta-) version for download requires self-signing the add-on through
 Find the API credentials in 1Password and run:
 
 ```shell
-WEB_EXT_API_KEY=xxx WEB_EXT_API_SECRET=xxx yarn build:firefox && yarn sign
+yarn build:firefox && WEB_EXT_API_KEY=xxx WEB_EXT_API_SECRET=xxx yarn sign
 ```
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -1,0 +1,13 @@
+# Maintainers 👔
+
+Tips & tricks for maintainers of Tickety-Tick.
+
+## Sign add-on for Firefox
+
+Offering a (beta-) version for download requires self-signing the add-on through [addons.mozilla.org](https://extensionworkshop.com/documentation/publish/#distribute-your-signed-extension). This generates a `.xpi` for installing the extension in Firefox.
+
+Find the API credentials in 1password and run:
+
+```
+yarn build:firefox && yarn sign 
+```
--- a/README.md
+++ b/README.md
@@ -1,4 +1,6 @@
-# Tickety-Tick [![Build Status](https://circleci.com/gh/bitcrowd/tickety-tick.svg?style=svg)](https://circleci.com/gh/bitcrowd/tickety-tick)
+# ![ticket icon](./src/common/icons/icon-32.png) Tickety-Tick
+
+[![Build Status](https://circleci.com/gh/bitcrowd/tickety-tick.svg?style=svg)](https://circleci.com/gh/bitcrowd/tickety-tick)
 
 > #### How do you name this branch? What is the message for that commit?
 > A browser extension to generate these for you, based on the ticket you're working on.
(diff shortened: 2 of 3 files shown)


MAINTAINERS.md
```diff
@@ -0,0 +1,13 @@
...
+Tips & tricks for maintainers of Tickety-Tick.
+
+## Sign add-on for Firefox
+
+Offering a (beta-) version for download requires self-signing the add-on through [addons.mozilla.org](https://extensionworkshop.com/documentation/publish/#distribute-your-signed-extension). This generates a `.xpi` for installing the extension in Firefox.
+
+Find the API credentials in 1password and run:
+
+```
+yarn build:firefox && yarn sign
```
pmeinhardt, line 12: Will this ask for the credentials or how does it work?

MAINTAINERS.md
```diff
@@ -0,0 +1,13 @@
...
+
+Tips & tricks for maintainers of Tickety-Tick.
+
+## Sign add-on for Firefox
+
+Offering a (beta-) version for download requires self-signing the add-on through [addons.mozilla.org](https://extensionworkshop.com/documentation/publish/#distribute-your-signed-extension). This generates a `.xpi` for installing the extension in Firefox.
+
+Find the API credentials in 1password and run:
+
+```
```
pmeinhardt, line 11:
```suggestion
```shell
```

MAINTAINERS.md
```diff
@@ -0,0 +1,13 @@
+# Maintainers 👔
+
+Tips & tricks for maintainers of Tickety-Tick.
+
+## Sign add-on for Firefox
+
+Offering a (beta-) version for download requires self-signing the add-on through [addons.mozilla.org](https://extensionworkshop.com/documentation/publish/#distribute-your-signed-extension). This generates a `.xpi` for installing the extension in Firefox.
+
+Find the API credentials in 1password and run:
```
pmeinhardt, line 9:
```suggestion
Find the API credentials in 1Password and run:
```
…is the official spelling.

README.md
```diff
@@ -86,7 +88,15 @@ Navigate to the [chrome://extensions](//chrome://extensions) page, enable "Devel
 
 If you just want to try out and debug the extension, go to [about:debugging#addons](//about:debugging#addons). Then press "Load Temporary Add-On" and select the built `manifest.json` from the `dist/firefox` extension directory.
 
-If you want to install this addon permanently, go to [about:addons](//about:addons) and click on the small cog icon. Select `Install Add-on From File...` and choose `dist/firefox`.
+If you want to install the extension permanently, you have two options:
+
+1. Install a self-signed add-on
+
+  Get a `tickety_tick-<version>.xpi` file from the [releases](https://github.com/bitcrowd/tickety-tick/releases) or [signing by a maintainer](./MAINTAINERS.md#sign-add-on-for-firefox). Then go to [about:addons](//about:addons), click on the small cog icon, select `Install Add-on From File...` and choose your `.xpi` file.
```
pmeinhardt, line 95:
```suggestion
  Get a `tickety_tick-<version>.xpi` file from the [releases](https://github.com/bitcrowd/tickety-tick/releases) or sign a build yourself as per our [maintainer guide](./MAINTAINERS.md#sign-add-on-for-firefox). Then go to [about:addons](//about:addons), click on the small cog icon, select `Install Add-on From File…` and choose your `.xpi` file.
```

pmeinhardt, line 95: We don't offer the `.xpi` files in the releases yet, so the documentation might mislead people in the meantime? Also, this doesn't really help me if I want to permanently install a custom-built version myself right? I'd still only be able to choose from the official releases (unless I'm a Tickety-Tick maintainer with access to the credentials).

So all in all, it seems to me like this is not really an option and everyone who wanted to tweak something about Tickety-Tick and use their own version can only use the 2. option?

🤔

klappradla: True for "tweak something and use their own version" but not for "I need to downgrade because the current version does not work for me" - which is a valid and in my opinion also very likely use case.

Mozilla's add-ons don't allow users to actually "control" which version they are using. Some people even stop semver-ing their extensions and instead use current date for the version 😂.

From what I see, installing from file is the only option to downgrade then 🤔 .

Using Chrome, I can download all old `.xps` files, so we could backfill at least the last two versions in the releases before moving to version 4 😉. Just to play safe.

pmeinhardt: Then let's rather create a separate section for "Installing an older version" and move the hint about the `xpi`/`xps` files there?

The "Install a custom-build version" section can then just focus on the "Enable unsigned add-ons" part.

I think this'd be less confusing. Depending on what I want to achieve, I can get clear separate instructions like this.

🙂

MAINTAINERS.md
klappradla: I'd provide them via environment variables. The command in 1Password looks like this:
```
WEB_EXT_API_KEY='xxx' WEB_EXT_API_SECRET='xxx' yarn build:firefox && yarn sign
```

I'll update the example to ☝️ for clarity.

README.md
klappradla: Makes sense 👍

klappradla: Moved it to a separate optional section in the installation instructions:

Screenshot 2020-05-07 at 10 43 44

klappradla:
Thanks for the helpful review @pmeinhardt 💚 

I applied all your suggestions 🏓

MAINTAINERS.md
```diff
@@ -0,0 +1,13 @@
...
+Tips & tricks for maintainers of Tickety-Tick.
+
+## Sign add-on for Firefox
+
+Offering a (beta-) version for download requires self-signing the add-on through [addons.mozilla.org](https://extensionworkshop.com/documentation/publish/#distribute-your-signed-extension). This generates a `.xpi` for installing the extension in Firefox.
+
+Find the API credentials in 1Password and run:
+
+```shell
+WEB_EXT_API_KEY=xxx WEB_EXT_API_SECRET=xxx yarn build:firefox && yarn sign
```
pmeinhardt, line 12: I think you'll probably want to set the environment variables on the second command.

```suggestion
yarn build:firefox && WEB_EXT_API_KEY=xxx WEB_EXT_API_SECRET=xxx yarn sign
```

To illustrate:

```sh
~ ❯ cat > try 
#!/usr/bin/env sh
echo $VAR
^C

~ ❯ chmod a+x try

~ ❯ VAR=yep ./try && ./try
yep

```

You should see 2 "yep"s here.

klappradla: 🤹 nice! I already had a vague feeling it may behave like that.

PR #239, opened 2020-05-08
[#210] Drop support for Safari

### LINKED ISSUE #210: Drop support for Safari
As it turns out, Safari dropped support for "Legacy Extensions" (see [release notes](https://developer.apple.com/documentation/safari_release_notes/safari_13_release_notes)). This means our current way of loading the TicketyTick extension in Safari via the `Extension Builder` is no longer possible from **v13** on - the current version of Safari.

There are new ways to [build an extension](https://developer.apple.com/documentation/safariservices/safari_app_extensions/building_a_safari_app_extension) and even ways to [convert a legacy extension](https://developer.apple.com/documentation/safariservices/safari_app_extensions/converting_a_legacy_safari_extension_to_a_safari_app_extension), but they all require diving into the world of Xcode 🕳  - and we probably cannot maintain this on the long run.

For some more insights, also see [this discussion](https://apple.stackexchange.com/questions/370551/show-extension-builder-absent-in-develop-menu-in-safari-13).

pmeinhardt:
Removes the code, build configuration and most of the indirections added to maintain a Safari version of Tickety-Tick. It's mostly deletions and renames (to get rid of the extra nesting and separation in the directory structure). The "biggest" change/diff is the merged Webpack config, which combines the previous `webpack.common.js` and `webpack.webext.babel.js`.

Closes #210.

✌️


src/options/index.jsx
```diff
@@ -1,4 +1,4 @@
-import './options.scss';
```
klappradla, line 1: 💚

PR #242, opened 2020-05-10
Upgrade Jest and Babel packages, clean up config

pmeinhardt:
- Remove unnecessary JSDOM from test setup code
- Upgrade Jest and Babel packages


pmeinhardt:
> Why can we get rid of the globals? Because we always assign / mock them right where tests need them? Or are they a leftover from earlier days?

Yeah, not sure anymore why we had it in there. A left-over for some reason. Jest already comes with the `"jsdom"` test environment as its default and we could even switch to the `"node"` one for a marginal speed-up since we're not mounting any components in the tests and explicitly instantiate (JS)DOMs for some of our test cases.

PR #246, opened 2020-05-12
Test initialization code

pmeinhardt:
I never really got around to testing this part of the codebase and was also wondering whether it'd be worth the effort. There are a few things going on in there though, so maybe it's a good idea?

I'm not 100% sure the way the tests are written is intuitive, so I am happy to hear about any thoughts you might have.

💚

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -11,7 +11,7 @@ module.exports = {
   snapshotSerializers: ['enzyme-to-json/serializer'],
   transform: {
     '^.+\\.jsx?$': 'babel-jest',
-    '^.+\\.png$': '<rootDir>/test/transforms/file.js',
+    '(?!^.+\\.jsx?$)': '<rootDir>/test/transforms/file.js',
   },
   transformIgnorePatterns: ['node_modules/(?!ky)'],
 };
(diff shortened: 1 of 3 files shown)


jest.config.js
```diff
@@ -11,7 +11,7 @@ module.exports = {
   snapshotSerializers: ['enzyme-to-json/serializer'],
   transform: {
     '^.+\\.jsx?$': 'babel-jest',
-    '^.+\\.png$': '<rootDir>/test/transforms/file.js',
+    '(?!^.+\\.jsx?$)': '<rootDir>/test/transforms/file.js',
```
klappradla, line 14: 🤷

src/popup/index.js
```diff
@@ -22,6 +20,7 @@ function close() {
 }
 
 async function load() {
+  const background = browser.extension.getBackgroundPage();
```
klappradla, line 23: 👍 makes sense as far as I can judge

jest.config.js
pmeinhardt: The `?!` is a so-called "negative lookahead". This'll match only if the negated expression does not match. If you look at what's inverted, it's exactly the `RegExp` on the preceding line. So this should match whenever the line above doesn't.

PR #251, opened 2020-06-08
Adjust browser-extension-page links

lwassermann:
because they are currently opened as `https://chrome/extensions`, which is not a valid url


lwassermann:
Hm. This does not work either...

PR #262, opened 2020-08-01
Upgrade dependencies

pmeinhardt:
✨

PR #263, opened 2020-08-01
[#257] Fix "back" chevron color

### LINKED ISSUE #257: Fix "back" chevron color
The chevron in the "back" button is black rather than the same color as the text:

screenshot

pmeinhardt:
Closes #257.

~Based on #262.~

The chevron color is fixed.

chevron

Also, the octicons package is upgraded to the most recent version.

This means the icons now look slightly different (as in GitHub's new design).

icons in popup

icons in settings

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "zip-webpack-plugin": "3.0.0"
   },
   "dependencies": {
-    "@githubprimer/octicons-react": "^8.5.0",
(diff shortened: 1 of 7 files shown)


pmeinhardt:
Should we mention the icon upgrade in the changelog?

pmeinhardt:
> Should we mention the icon upgrade in the changelog?

Just noticed we don't really have a changelog, so nevermind. 😁

PR #264, opened 2020-08-02
[#261] Support "dark mode" 

### LINKED ISSUE #261: Support "dark mode"
It'd be cool to make the Tickety-Tick popup and preference page aware of what the browser/OS color scheme preference is and to adjust the background and text colors.

This could possibly achieved through the [`prefers-color-scheme`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) CSS media feature.

pmeinhardt:
Closes #261.

popup

To do:

- [x] use light color version of logo on about page in dark mode?
- [ ] ~dark theme for options page?~ (not themed for now)
- [x] clean up


pmeinhardt:
Unfortunately Bootstrap does not support customization through CSS variables. It exports theme info (colors…) as CSS variables, but does not read variables to allow adjusting styling. The current solution therefore instead of tweaking variables based on media queries, adjusts the CSS classes.

An alternative solution could be to build 2  CSS files with complete versions of Bootstrap - just with different theming - and to include those via `<link href="…" rel="stylesheet" media="(prefers-color-scheme: …)">`.

klappradla:
> The current solution therefore instead of tweaking variables based on media queries, adjusts the CSS classes.

Given the bootstrap-circumstances, this is still pretty close to the approach suggested in https://codeguide.co/#css-media-queries and perfectly fine in my opinion 👌

pmeinhardt:
Switched the logo on the help page to use the SVG and style it based on theme colors:

logo

pmeinhardt:
I ended up creating 2 themed versions of (a trimmed-down) Bootstrap, one of which is nested under `.dark` to conditionally apply these more specific styles if `(prefers-color-scheme: dark)` matches.

This of course about doubles the size of the CSS and adds a lot of duplication, but - hey 🤷 - it's about 186K now and it's not like we're transferring this over the network anyways.

src/popup/index-dark.scss
```diff
@@ -0,0 +1,5 @@
+.dark {
+  @import 'styles/settings/common';
+  @import 'styles/settings/dark';
+  @import 'styles/main';
```
klappradla, line 4: I may be missing something, but why does `.dark` also contain the `common` and `main` styles? I thought the `dark` part was coming on top of what's already there 🤔

klappradla:
> popup

Currently in the middle of my review, but I don't see the right place to ask this 🙈: why does the menu bar icon not change for you? This is a behavior for Firefox I already added in #227 🤔 

See here ⏬ 
Kapture 2020-08-05 at 10 08 12

src/popup/styles/settings/_dark.scss
```diff
@@ -0,0 +1,25 @@
+// Dark theme
+
+// Invert gray-scale colors
```
klappradla, line 3: Did you invert what was there, or are those the colors from [the styleguide](https://design.firefox.com/photon/visuals/iconography.html)?

pmeinhardt:
> Currently in the middle of my review, but I don't see the right place to ask this 🙈: why does the menu bar icon not change for you? This is a behavior for Firefox I already added in #227 🤔

The screenshot was taken in Chrome. In Firefox, the icon behaves as expected. The changes here don't touch the manifest. ✌️

klappradla:
> The screenshot was taken in Chrome. In Firefox, the icon behaves as expected. The changes here don't touch the manifest. ✌️

On my 🙈 Time for holidays and sorry for the noise 🕳️

src/popup/observe-media.js
```diff
@@ -0,0 +1,5 @@
+export default function observe(specifier, fn) {
```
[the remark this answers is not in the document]
pmeinhardt: Not sure whether we can control the menubar icon, as it's statically specified in the manifest as far as I remember.

src/popup/index-dark.scss
pmeinhardt: common is just the Bootstrap Sass variables that I want to apply to both the default and the dark theme. It essentially just makes sure, we start out with the same values (Sass processing is somewhat weird).

src/popup/styles/settings/_dark.scss
pmeinhardt: I inverted what was there in Bootstrap, as we're using that for the default theme and browsers other than Firefox. 🤔 I only adjusted the background color to be a more warm shade of gray which aligns better with what both Firefox and Chrome use.

src/popup/observe-media.js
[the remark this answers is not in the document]
klappradla: One cannot control it from "within" the extension, it requires some extra JS listeners in the `content/background.js`. I just meant if we already have some JS for changing this in place, then having the listeners for the icon isn't too far of either (see e.g. [this post](https://stackoverflow.com/questions/58880234/toggle-chrome-extension-icon-based-on-light-or-dark-mode-browser)).

But let's not open this can of worms here.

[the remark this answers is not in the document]
pmeinhardt: Uff, yeah. Maybe another time 😀

PR #271, opened 2020-10-05
[#270] GitLab integration is broken

### LINKED ISSUE #270: GitLab integration is broken
Hey bitcrowd 👋

Looks like GitLab removed the `js-issuable-app-initial-data` id from the issue page causing tickety-tick to fail with the error below:

Tickety-Tick revision: fda1749a21ac75aa115d6245fac21e63570b16c3

```
Cannot read property 'innerHTML' of null
    at scan (webpack:///src/core/adapters/gitlab.js:13:20)
    at attempt (webpack:///src/core/search.js:10:26)
    at search (webpack:///src/core/search.js:28:47)
    at search (webpack:///src/core/search.js:28:25)
    at browser.runtime.onMessage.addListener( (webpack:///src/content/index.js:8:13)
    at wrappedSendResponse (webpack:///node_modules/webextension-polyfill/dist/browser-polyfill.js:1072:21)
```

Taig:
Implements a new approach to extract the ticket id from a GitLab issue page. The previously used id `js-issuable-app-initial-data` is no longer present in the source. However the legacy approach is left in place to support older self-hosted GitLab installations.

I tested the development build with Chrome on the GitLab page to confirm it's working.

Closes #270

### DIFF
--- a/src/core/adapters/gitlab.js
+++ b/src/core/adapters/gitlab.js
@@ -7,19 +7,21 @@
 import { $has, $text } from './helpers';
 
 function findTicketId(doc) {
-  const initialDataEl = doc.getElementById('js-issuable-app-initial-data');
+  const ticketId = doc.body.dataset.pageTypeId;
 
-  if (initialDataEl !== null) {
-    // Legacy approach of extracting the ticket id, left in place to support
-    // older self-hosted GitLab installations
-    const initialData = JSON.parse(
-      initialDataEl.innerHTML.replace(/&quot;/g, '"')
-    );
-
-    return initialData.issuableRef.match(/#(\d+)/)[1];
+  if (ticketId) {
+    return ticketId;
   }
 
-  return doc.body.dataset.pageTypeId;
+  const initialDataEl = doc.getElementById('js-issuable-app-initial-data');
+
+  // Legacy approach of extracting the ticket id, left in place to support
+  // older self-hosted GitLab installations
+  const initialData = JSON.parse(
+    initialDataEl.innerHTML.replace(/&quot;/g, '"')
+  );
+
+  return initialData.issuableRef.match(/#(\d+)/)[1];
 }
 
 async function scan(loc, doc) {
--- a/src/core/adapters/gitlab.js
+++ b/src/core/adapters/gitlab.js
@@ -6,14 +6,25 @@
 
 import { $has, $text } from './helpers';
 
-async function scan(loc, doc) {
-  if (doc.body.dataset.page === 'projects:issues:show') {
-    const initialDataEl = doc.getElementById('js-issuable-app-initial-data');
+function findTicketId(doc) {
+  const initialDataEl = doc.getElementById('js-issuable-app-initial-data');
+
+  if (initialDataEl !== null) {
+    // Legacy approach of extracting the ticket id, left in place to support
+    // older self-hosted GitLab installations
     const initialData = JSON.parse(
       initialDataEl.innerHTML.replace(/&quot;/g, '"')
     );
 
-    const id = initialData.issuableRef.match(/#(\d+)/)[1];
+    return initialData.issuableRef.match(/#(\d+)/)[1];
+  }
+
+  return doc.body.dataset.pageTypeId;
+}
+
+async function scan(loc, doc) {
+  if (doc.body.dataset.page === 'projects:issues:show') {
+    const id = findTicketId(doc);
     const title = $text('.issue-details .title', doc);
     const type = $has('.labels [data-original-title="bug"]', doc)
       ? 'bug'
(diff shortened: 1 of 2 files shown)


src/core/adapters/gitlab.test.js
```diff
@@ -3,6 +3,18 @@ import { JSDOM } from 'jsdom';
...
+      <div class="issue-details">
+        <div class="detail-page-description">
+          <h2 class="title">A Random GitLab Issue</h2>
+        </div>
+      </div>
+    </div>
+  </body>
+`;
+
+const LEGACYISSUE = `
```
Taig, line 17: I couldn't name this `LEGACYISSUEPAGE`. It seemed like things started breaking after a certain name length 🤷‍♂️

klappradla:
Yay, thanks a ton for jumping onto this @Taig 👏 

I just briefly checked your implementation on `gitlab.com` and a self hosted version. Looking good 💯

src/core/adapters/gitlab.js
```diff
@@ -6,14 +6,25 @@
 
 import { $has, $text } from './helpers';
 
-async function scan(loc, doc) {
-  if (doc.body.dataset.page === 'projects:issues:show') {
-    const initialDataEl = doc.getElementById('js-issuable-app-initial-data');
+function findTicketId(doc) {
+  const initialDataEl = doc.getElementById('js-issuable-app-initial-data');
```
klappradla, line 10: Purely cosmetical, but maybe it would be easier to reason about this for future developers, if we turned this around: first check for the _"new"_ structure and then have an optional fallback for the legacy approach?

This way, it could eventually also be grouped more concise into one "current" and one "legacy" block, which would make it easier to get rid of the legacy logic at a later point in time.

Maybe roughly something along the lines of this:
```js
const ticketId = doc.body.dataset.pageTypeId;

if (ticketId) {
  return ticketId;
} else {
  // Legacy approach for older self-hosted GitLab instances
  // ...
```

klappradla (commented):
Looking good in my opinion 💅 

I left a small cosmetic comment - let me know what you think.

Aside this: did you have the chance to test this on "older" instances as well? The only self hosted instance I have access to also has the data attribute set on the `body` already. Therefore: do we even know if it's necessary to support the legacy approach any more? Or is this targeting a version of Gitlab which pretty much no one is using any more?

Taig:
> Therefore: do we even know if it's necessary to support the legacy approach any more? Or is this targeting a version of Gitlab which pretty much no one is using any more?

The body attribute was probably introduced [10 months ago](https://gitlab.com/gitlab-org/gitlab/-/commit/cb3f40c95d6f6973dffb376ae861dfa31a7fded7). I clicked my way through 10 [popular GitLab instances](https://wiki.p2pfoundation.net/List_of_Community-Hosted_GitLab_Instances), all of which had a recent enough installation except for [Mastodon](https://source.joinmastodon.org/explore/projects/). It's up to you to decide whether the legacy support can already be dropped. But I also don't mind making the changes to the code that you suggested.

src/core/adapters/gitlab.js
Taig: The linter didn't allow me to write the if statement with an `else` clause because of the early return. That's when I started to curse at the computer and just do whatever pleases the linter. But I'll take another look at it.

klappradla:
> The body attribute was probably introduced [10 months ago](https://gitlab.com/gitlab-org/gitlab/-/commit/cb3f40c95d6f6973dffb376ae861dfa31a7fded7). I clicked my way through 10 [popular GitLab instances](https://wiki.p2pfoundation.net/List_of_Community-Hosted_GitLab_Instances), all of which had a recent enough installation except for [Mastodon](https://source.joinmastodon.org/explore/projects/).

Wow, what an awesome research 💯 

Then let's stay with the legacy support for now 👍

src/core/adapters/gitlab.js
klappradla: Oh, good point 💡 

I guess the linter probably wants code like this (simply skipping the `else`):
```js
const ticketId = doc...;

if (ticketId) {
  return ticketId;
}
// Legacy support
const otherId = ...
return ...
```

src/core/adapters/gitlab.js
```diff
@@ -6,14 +6,27 @@
...
+  const ticketId = doc.body.dataset.pageTypeId;
+
+  if (ticketId) {
+    return ticketId;
+  }
+
+  const initialDataEl = doc.getElementById('js-issuable-app-initial-data');
+
+  // Legacy approach of extracting the ticket id, left in place to support
+  // older self-hosted GitLab installations
```
klappradla, line 19:
```suggestion
  // Legacy approach of extracting the ticket id, left in place to support
  // older self-hosted GitLab installations
  const initialDataEl = doc.getElementById('js-issuable-app-initial-data');
```
Would place this after the comment.

klappradla (approved):
Yay 👏 

Thanks for the updates! Good to go from my side - would only reformat the comment.

PR #274, opened 2020-11-14
Add tara ai support

schanso:
Hi everyone 👋 

Hope you're doing well! I recently started fiddling around with [tara.ai](https://app.tara.ai) for a side project and quickly realized manually keeping branch names and CLI commands in sync with what I've grown to love with `tickety-tick` was too painful. So I added a Tara adapter here.

I tried to follow practices used in other adapters closely and added documentation where I thought it was necessary.
___

### How it works 👷‍♀️ 

Tara uses workspaces with multiple requirements, or projects, each. Task IDs are 1-based indices unique within each workspace (__not__ within the requirement). Page URLs follow the pattern: `https://app.tara.ai/<workspace-name>/tasks/<id>`.

The adapter extracts ticket information from the DOM. Given the obfuscated CSS classes, most context is retrieved from `data-cy` attributes. This is obviously not ideal, given they might vanish from production code at some point. 😒 However, any other CSS changes would probably require the same need for maintenance?

Tara has two ways to present tickets,
1) in a modal-like overlay
tara_modal

2) in full-page view
tara_full-page

This is important, because they use `draft-js` for rich text editing, and 1. has an extra editor node that needs to be skipped in order to extract the ticket description.

**Example branch**
```
feature/task-8-add-some-feature
```

**Example message**
```
[#TASK-8] Add some feature

We are currently missing some feature in our app. Let's "change" this.
See `attached` screenshot for more information .

https://app.tara.ai/private-projects/tasks/8
```

**Example CLI command**
```
git checkout -b 'feature/task-8-add-some-feature' && git commit --allow-empty -m '[#TASK-8] Add some feature

We are currently missing some feature in our app. Let'\''s "change" this.
See `attached` screenshot for more information .

https://app.tara.ai/private-projects/tasks/8
'
```

Tara does not give the option to distinguish between story, bug, task etc., so the ticket `type` is hard-coded as `"feature"` for now.

___

### Demo 💿 
You can see the adapter in action here 👇 

tara_demo

___

Please let me know if you don't see the need for a Tara adapter right now or in this form. Your 👀 would be highly appreciated, of course.

I'll show myself out 💃

### DIFF
--- a/src/core/adapters/tara.js
+++ b/src/core/adapters/tara.js
@@ -1,13 +1,13 @@
 /**
  * tara.ai adapter
  *
- * The adapter extracts the current task's identifier from the page url
+ * The adapter extracts the current task's identifier from the page URL
  * and uses the DOM to retrieve information for the corresponding ticket.
- * Page urls follow the pattern:
+ * Page URLs follow the pattern:
  *
  *   https://app.tara.ai/<workspace-name>/tasks/<id>
  *
- * Each workspace has multiple projects (called "requirements"), task ids
+ * Each workspace has multiple projects (called "requirements"), task IDs
  * always start with `TASK-` followed by a 1-based index unique within the
  * given workspace.
  */
@@ -17,7 +17,7 @@ import { match } from 'micro-match';
 import { $all, $has, $text } from './helpers';
 
 /**
- * Tara has two ticket views, one in a modal-like overlay, one in full page view.
+ * Tara has two ticket views, one in a modal-like overlay, one in full-page view.
  */
 function isModalView(doc) {
   const modalViewSelector = 'input[data-cy="requirement-detail-name-input"]';
--- a/src/core/adapters/tara.js
+++ b/src/core/adapters/tara.js
@@ -25,17 +25,13 @@ function isModalView(doc) {
 }
 
 function extractDescription(doc) {
-  const description = [];
   const editorRootNodes = $all('.DraftEditor-root', doc);
 
   if (isModalView(doc)) editorRootNodes.shift(); // Skip project information editor node
 
   const textNodes = $all('span[data-text="true"]', editorRootNodes[0]);
-  textNodes.forEach((node) => {
-    description.push(node.textContent);
-  });
 
-  return description.join('\n');
+  return textNodes.map((node) => node.textContent).join('\n');
 }
 
 async function scan(loc, doc) {
--- a/src/core/adapters/tara.js
+++ b/src/core/adapters/tara.js
@@ -1,0 +1,57 @@
+/**
+ * tara.ai adapter
+ *
+ * The adapter extracts the current task's identifier from the page url
+ * and uses the DOM to retrieve information for the corresponding ticket.
+ * Page urls follow the pattern:
+ *
+ *   https://app.tara.ai/<workspace-name>/tasks/<id>
+ *
+ * Each workspace has multiple projects (called "requirements"), task ids
+ * always start with `TASK-` followed by a 1-based index unique within the
+ * given workspace.
+ */
+
+import { match } from 'micro-match';
+
+import { $all, $has, $text } from './helpers';
+
+/**
+ * Tara has two ticket views, one in a modal-like overlay, one in full page view.
+ */
+function isModalView(doc) {
+  const modalViewSelector = 'input[data-cy="requirement-detail-name-input"]';
+  return $has(modalViewSelector, doc);
+}
+
+function extractDescription(doc) {
+  const description = [];
+  const editorRootNodes = $all('.DraftEditor-root', doc);
+
+  if (isModalView(doc)) editorRootNodes.shift(); // Skip project information editor node
+
+  const textNodes = $all('span[data-text="true"]', editorRootNodes[0]);
+  textNodes.forEach((node) => {
+    description.push(node.textContent);
+  });
+
+  return description.join('\n');
+}
+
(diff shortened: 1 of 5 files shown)


src/core/adapters/tara.test.js
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  it('returns an empty array if it is on a different page', async () => {
+    const location = loc('some-other-website.com');
+    const result = await scan(location, dom());
+
+    expect(result).toEqual([]);
+  });
+
+  it('extracts tickets from task pages', async () => {
```
klappradla, line 38: Hm, would it also make sense to add another scenario for the modal view? If it differs and we already maintain a separate code path for it, it could be useful to also reflect this in the tests.

src/core/adapters/tara.js
```diff
@@ -0,0 +1,57 @@
...
+  if (loc.host !== 'app.tara.ai') return [];
+
+  const { index } = match('/:workspace/tasks/:index', loc.pathname);
+
+  if (!index) return [];
+
+  const id = $text('span[data-cy="task-modal-task-id-text"]', doc);
+  const description = extractDescription(doc);
+  const title = $text('textarea[data-cy="task-modal-task-name-text"]', doc);
+  const type = 'feature';
```
klappradla, line 51: Just outta curiosity: why did you decide to hardcode this to `feature`? I'd have expected it to be `task` if that's also their internal name.

klappradla (commented):
Hey @schanso 👋 

Nice to hear from you again 🤗 

And ace that you're jumping onto this 💯

I haven't yet heard of **Tara** - nor used it of course - but I'll give it a spin with your code as soon as I find the time. The adapter implementation looks neat from a first look ✨ 

Do you know if **Tara** is implemented as a single page application powered by an (probably internal) API under the hoods? If so, you could check in the network tab, if the communication with their API could also be a useful source for Tickety Tick 🕵️.

Just a though... For other adapters - Jira and Notifion for instance - using the API was easier to maintain in contrast to relying on extracting information directly from the DOM. Maybe this could also be a useful option here 🤔

nickgnd:
I passed by to say hello to @schanso, nice to see you here (and great work, thanks :bow)

src/core/adapters/tara.js
```diff
@@ -0,0 +1,57 @@
...
+  const editorRootNodes = $all('.DraftEditor-root', doc);
+
+  if (isModalView(doc)) editorRootNodes.shift(); // Skip project information editor node
+
+  const textNodes = $all('span[data-text="true"]', editorRootNodes[0]);
+  textNodes.forEach((node) => {
+    description.push(node.textContent);
+  });
+
+  return description.join('\n');
```
pmeinhardt, line 38: I think you should be able to just use:

```javascript
return textNodes.map((node) => node.textContent).join('\n');
```

pmeinhardt (commented):
Thanks a ton @schanso 💚 The code looks very neat & clean. ✨ 

With a few small adjustments (like those @klappradla already pointed out) I'd be very happy if we can merge this. 🙌 

I hope you're doing alright? 🙂

src/core/adapters/tara.js
pmeinhardt: We've been using `feature` as a default so far if we can't extract a specific "type" (as per our README).

readme

I think this was inspired by the "feature-branching" conventions. Not sure whether this convention still makes sense or whether we should go with `task` here and remove the "(default)" from the README? @klappradla? @andreasknoepfle? @nickgnd?

pmeinhardt (commented):
As @klappradla pointed out, if Tara has an API that you can query from the page context this could make the adapter more reliable. This could be something to tackle in a possible follow-up PR though if needed. 🙂

schanso:
Thanks for coming back to this so quickly! 😊 I'm doing well, but glancing at your avatars and reading your comments throws me back to when life was a little easier on all of us.

> communication with their API could also be a useful source for Tickety Tick

That was my initial hope, too, and there is a lot going on in the network tab, but mostly analytics. I couldn't deduce anything useful, unfortunately.

> With a few small adjustments

I will get to them, hopefully tonight. Thanks for the reviews, @klappradla  and @pmeinhardt  💪 

> I passed by to say hello to @schanso

Thanks for stopping by and the kind words! ❤️

src/core/adapters/tara.js
schanso: Plus, it felt like an odd repetition, given that `task` was already part of the `id` itself, i. e. `task/task-8`. I'd be happy to change this to something more expressive, but I couldn't come up with anything clever yet.

schanso:
I worked in your feedback, @klappradla and @pmeinhardt, please have another look whenever you find the time 🙇

pmeinhardt (approved):
Neat 🙌  I'd be happy to merge this. 🙂

@klappradla, maybe you can also take another look on Monday. ✌️ 
Then, once merged, maybe we can push a new release on the Wednesday HACKNIGHT™️. 👾 

Awesome work @schanso! 💚

src/core/adapters/tara.js
klappradla: > We've been using `feature` as a default so far if we can't extract a specific "type" (as per our README).

Perfectly fine 👍 

I tend to call everything I'm doing "chore", but that's probably just because I a) like the word while not knowing exactly what it means and b) because the work I'm doing feels more like 🚜

PR #277, opened 2020-12-07
Add link to open extension settings to popup

pmeinhardt:
During a pair-programming session with @nickgnd, I noticed that it takes a little too long to navigate to the extension settings when you just want to quickly make some adjustments.

This adds a link to the popup to open the options page. ✌️ 

preview

### DIFF
--- a/src/popup/components/tool.jsx
+++ b/src/popup/components/tool.jsx
@@ -1,7 +1,8 @@
 import PropTypes from 'prop-types';
-import React from 'react';
+import React, { useContext } from 'react';
 import { Link } from 'react-router-dom';
 
+import EnvContext from '../env-context';
 import ErrorShape from '../utils/error-shape';
 import TicketShape from '../utils/ticket-shape';
 import Content from './content';
@@ -10,13 +11,27 @@ import NoTickets from './no-tickets';
 import TicketControls from './ticket-controls';
 
 function Tool({ tickets, errors }) {
+  const { close, openopts } = useContext(EnvContext);
+
+  async function onClickSettingsLink(event) {
+    event.preventDefault();
+    await openopts();
+    close();
+  }
+
   return (
     <>
       <Navbar>
         <span className="navbar-brand">Tickety-Tick</span>
         <ul className="navbar-nav ml-auto">
           <li className="nav-item">
-            <Link className="nav-link" to="/about">
+            {/* eslint-disable-next-line jsx-a11y/anchor-is-valid */}
+            <a className="nav-link" href="#" onClick={onClickSettingsLink}>
+              Settings
+            </a>
+          </li>
+          <li className="nav-item">
+            <Link className="nav-link pr-0" to="/about">
               Help
             </Link>
           </li>
--- a/src/popup/components/tool.jsx
+++ b/src/popup/components/tool.jsx
@@ -25,8 +25,11 @@ function Tool({ tickets, errors }) {
         <span className="navbar-brand">Tickety-Tick</span>
         <ul className="navbar-nav ml-auto">
           <li className="nav-item">
-            {/* eslint-disable-next-line jsx-a11y/anchor-is-valid */}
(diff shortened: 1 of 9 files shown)


src/popup/components/tool.jsx
```diff
@@ -10,13 +11,27 @@ import NoTickets from './no-tickets';
 import TicketControls from './ticket-controls';
 
 function Tool({ tickets, errors }) {
+  const { close, openopts } = useContext(EnvContext);
+
+  async function onClickSettingsLink(event) {
+    event.preventDefault();
+    await openopts();
+    close();
+  }
```
nickgnd, line 20: 👍

nickgnd (approved):
That's a great idea!! 💯

src/popup/components/tool.jsx
```diff
@@ -10,13 +11,27 @@ import NoTickets from './no-tickets';
 import TicketControls from './ticket-controls';
 
 function Tool({ tickets, errors }) {
+  const { close, openopts } = useContext(EnvContext);
+
+  async function onClickSettingsLink(event) {
```
klappradla, line 16: Just outta curiosity: is this a React / JS naming pattern, to use `on<action-name><content>` for these functions? I'd personally have gone with something that indicates what the function does, rather than were it's placed: `openSettingsPage` for instance.

src/popup/styles/settings/_common.scss
```diff
@@ -10,3 +10,4 @@ $primary: #c32f34;
 
 $navbar-brand-font-size: $font-size-base;
 $nav-link-padding-y: 0;
+$navbar-nav-link-padding-x: 0.25rem;
```
klappradla, line 13: I don't see this being used 🔍

klappradla (approved):
Awesome idea 💯

src/popup/components/tool.jsx
```diff
@@ -10,13 +11,27 @@ import NoTickets from './no-tickets';
...
+
   return (
     <>
       <Navbar>
         <span className="navbar-brand">Tickety-Tick</span>
         <ul className="navbar-nav ml-auto">
           <li className="nav-item">
-            <Link className="nav-link" to="/about">
+            {/* eslint-disable-next-line jsx-a11y/anchor-is-valid */}
+            <a className="nav-link" href="#" onClick={onClickSettingsLink}>
```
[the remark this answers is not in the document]
pmeinhardt: Unfortunately the `.nav-link` style does not properly style buttons. 😞 I didn't want to go down a 🐇🕳️ to try and make them work. I'll have another look though whether using `.btn-link` in combination with `.nav-link` can trick a button into looking good there. ✌️

src/popup/components/tool.jsx
pmeinhardt: Since this function is specifically a click handler calling `event.preventDefault()`, using this slightly more verbose and specific name seemed appropriate to me. I'm not sure whether there is an "official" convention here, but I've definitely seen similar naming before.

src/popup/styles/settings/_common.scss
pmeinhardt: It's a variable used by the [Bootstrap SCSS](https://github.com/twbs/bootstrap/blob/v4.5.3/scss/_variables.scss#L721).

klappradla: 💡 oh, thanks!

src/popup/components/tool.jsx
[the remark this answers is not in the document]
pmeinhardt: Tried with making it a button and applying the `btn` and `btn-link` classes, but that still makes them look and behave different from the help link then.

Instead, I opted for updating the `href` attribute to point to `options.html`. This is actually valid and if you didn't have the `onClick` handler, it would look like this: 😀 

options page in popup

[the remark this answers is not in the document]
klappradla: Haha, would have expected this to work 😂

PR #278, opened 2020-12-08
Use a package for copying to the clipboard

pmeinhardt:
Use [`copy-text-to-clipboard`](https://github.com/sindresorhus/copy-text-to-clipboard) for copying to the clipboard instead of a custom implementation. This simplifies our code and the package takes care of some extra handling. :v:


klappradla (approved):
Nice 👏

PR #281, opened 2021-01-12
Upgrade dependencies

pmeinhardt:
- Update packages that depend on node-notifier
- Upgrade build dependencies (minor- and patch-level versions)
- Upgrade build dependencies (major versions)
- Upgrade runtime dependencies
- Upgrade node version


tessi (approved):
✨ thanks for the updates @pmeinhardt 💚

PR #287, opened 2021-02-18
Extracted formatter into its own package

tessi:
I extracted the formatter logic into its own package at
https://github.com/bitcrowd/tickety-tick-formatter

The reason is that we often have project ideas that are not
browser-plugins, but still want to format tickets into git-digestable
strings.

One example is the GitHub integration for OpenProject I am currently
working on.

A beneficial side effect is, that the formatter/template logic is
extracted and now implemented in TypeScript.


src/options/components/form.test.jsx
```diff
@@ -153,8 +153,8 @@ describe('form', () => {
     const wrapper = render({});
     const text = wrapper.text();
 
-    Object.values(helpers).forEach((fn) => {
-      expect(text).toContain(fn.name || fn.description);
+    Object.keys(helpers).forEach((helperName) => {
+      expect(text).toContain(helpers[helperName].description || helperName);
```
pmeinhardt, line 157:
```suggestion
    Object.entries(helpers).forEach(([name, fn]) => {
      expect(text).toContain(fn.description || name);
```

pmeinhardt (commented):
We’ve been at this crossroad – “extract or  not” – a few times already and this looks like the time to go ahead and extract. :green_heart: :thumbsup:

The main reason we haven't done it in the past is that it will make updating Tickety-Tick slightly more work. It means we'll have to make changes in the formatter repo, ~publish a new version of the package~ (does not apply if we pull the package from the repo) and then create a PR here. On the other hand, the formatting has been relatively stable since we introduced the commit message formatting. 🤔

package.json
```diff
@@ -80,17 +80,15 @@
...
     "react": "17.0.1",
     "react-dom": "17.0.1",
     "react-router": "5.2.0",
     "react-router-dom": "5.2.0",
     "react-svg": "^11.2.2",
     "react-textarea-autosize": "^8.3.0",
-    "speakingurl": "14.0.1",
     "stacktrace-js": "^2.0.0",
-    "strip-indent": "^3.0.0",
+    "tickety-tick-formatter": "bitcrowd/tickety-tick-formatter",
```
pmeinhardt, line 91: 💡 I hadn't yet seen that you can now use such shortcuts for GitHub URLs.

— https://docs.npmjs.com/cli/v7/configuring-npm/package-json#github-urls

pmeinhardt (approved):

PR #292, opened 2021-04-15
[291] Convert Jira markup to markdown

### LINKED ISSUE #291: Convert Jira markup to markdown
👋  I would really like to be able to take the description from Jira tickets and copy it to commit messages.

maltoe:
Fixes #291

### DIFF
--- a/src/core/adapters/jira.js
+++ b/src/core/adapters/jira.js
@@ -46,20 +46,21 @@ function getSelectedIssueId(loc, prefix = '') {
 function extractTicketInfo(response, host) {
   const {
     key: id,
-    fields: { issuetype, summary: title, description },
+    fields: { issuetype, summary: title, description: descriptionJira },
   } = response;
   const type = issuetype.name.toLowerCase();
   const url = `https://${host}/browse/${id}`;
 
-  const descriptionMd =
-    description && j2m.to_markdown(description).replace('|smart-link', '');
+  const description =
+    descriptionJira &&
+    j2m.to_markdown(descriptionJira).replace('|smart-link', '');
 
   return {
     type,
     id,
     title,
+    description,
     url,
-    description: descriptionMd,
   };
 }
 
--- a/src/core/adapters/jira.js
+++ b/src/core/adapters/jira.js
@@ -12,6 +12,7 @@
  * - Issue view: https://<YOUR-SUBDOMAIN>.atlassian.net/browse/<ISSUE-KEY>
  */
 
+import j2m from 'jira2md';
 import { match } from 'micro-match';
 
 import client from '../client';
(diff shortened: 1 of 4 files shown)


src/core/adapters/jira.js
```diff
@@ -50,12 +51,15 @@ function extractTicketInfo(response, host) {
...
+  const descriptionMd =
+    description && j2m.to_markdown(description).replace('|smart-link', '');
+
   return {
     type,
     id,
     title,
-    description,
     url,
+    description: descriptionMd,
```
pmeinhardt, line 62: It might be nice to rename the `const` to a slightly less cryptic `markdownDescription` or alternatively renaming the `description` from the `response` to `jiraText` or `jiraDescription` and using the plain `description` for the converted markdown version.

pmeinhardt (approved):
Nice 💖 Great to see that somebody seems to have picked up the previously dormant/unmaintained `jira2md` package. 🥳

pmeinhardt:
Maybe it's worth testing this out with a local build for another week or so before we release? 🤔 It somehow expect the JIRA text to markdown to be somewhat tricky and would like to gain some confidence in the new package this PR introduces to make sure we don't unintentionally break Tickety-Tick for JIRA users.

maltoe:
> a local build for another week or so

@pmeinhardt running with a local build right now, can tell you in a week how it goes

maltoe:
@pmeinhardt Found the first couple oddities:

* If you have an apostrophe in your test (`don't`), Jira will turn it into an actual english apostrophe (`don’t`) which the converter lets through -> I guess from the converter perspective this is intended, but I'd like it if it converted them to simple quotes instead. Maybe that's not an issue with this PR though, but could be solved in a more general way?
* I had a line `Let’s unify them into one!` in a ticket in Jira which came out as `Let’s unify them into one\!` 🤔 
* The converter turned `{{assert_count_difference}}` (monospace expression) into `[backtick]assert*count*difference[backtick]` (with `[backtick]` being actual backticks) :(

maltoe:
Sometimes it also copies a whole bunch of bogus text with a lot of repetitions in it. Looks like `jira2md` is broken after all. :/

maltoe:
@pmeinhardt I give up, have disabled the description in my config. It's entirely broken for any reasonable real-world description. The last point I mentioned with the repetitions is the worst, at times it generates so much garbage that it doesn't even fit the command line anymore.

pmeinhardt:
> @pmeinhardt I give up, have disabled the description in my config. It's entirely broken for any reasonable real-world description. The last point I mentioned with the repetitions is the worst, at times it generates so much garbage that it doesn't even fit the command line anymore.

Oh wow. I had a hunch that the Atlassian markup might become a challenge, but I wonder what's going on in the conversion package that makes it "generate so much garbage". 🤔

Maybe we can explore some of the other possible approaches in an upcoming hacknight. 👾

PR #293, opened 2021-04-21
Update dependencies

pmeinhardt:


webpack.config.babel.js
```diff
@@ -142,7 +142,7 @@ config.plugin('copy').use(CopyWebpackPlugin, [
     patterns: [
       {
         from: src('icons', '*.png'),
-        to: '[name].[ext]',
+        to: '[name][ext]',
```
pmeinhardt, line 145: https://github.com/webpack-contrib/copy-webpack-plugin/blob/master/CHANGELOG.md#800-2021-03-04

pmeinhardt:
Waddup Circle? 🤔

pmeinhardt:
Bumping the Circle CI cache key seems to not do the trick. Can somebody check this out locally, `yarn install` and `yarn test` to see whether it's an issue on my machine?

nickgnd:
@pmeinhardt I can reproduce the issue on my macbook

image

Apparently `commit` is undefined

> TypeError: Cannot read property 'commit' of undefined

And I guess it is due to this package `tickety-tick-formatter` that does not have any named export `defaults`

https://github.com/bitcrowd/tickety-tick-formatter/blob/main/src/index.ts

nickgnd:
@pmeinhardt I pushed a commit, it should fix the issue 😉

pmeinhardt:
> @pmeinhardt I can reproduce the issue on my macbook […] and I guess it is due to this package `tickety-tick-formatter` that does not have any named export `defaults`

Argh. Thanks @nickgnd. Something seems to be off with my local package installation then and I did not have the time this morning to properly investigate. 🙄 

Thanks a ton for the spot + fix. 💖

nickgnd (approved):
👍 Nice!

Thanks a ton 🙇

pmeinhardt:
@tessi even if we install `tickety-tick-formatter` from GitHub, it might be nice to tag specific commits with a semver version number so it's clear whether the public API has changed in the meantime or not.

In this case, the `defaults` export had changed to `templateDefaults`:
https://github.com/bitcrowd/tickety-tick-formatter/commit/b5c5b38f1f29ed5c839fba76baffc3589b526eac

PR #294, opened 2021-04-21
Fix format preview on options page

pmeinhardt:
**Before**

before

**After**

fixed


src/options/components/form.jsx
```diff
@@ -84,7 +84,15 @@ class Form extends Component {
   render() {
     const { loading, autofmt, ...templates } = this.state;
 
-    const fmt = format(templates, autofmt);
+    // Use default for preview if template is blank
```
klappradla, line 87: 💡

klappradla (approved):
Gold 💯

PR #298, opened 2021-05-01
Remove env context

pmeinhardt:
We no longer have different entry points for Safari and other browsers.
We can drop the formerly browser-specific `EnvContext` and simplify the
setup.

### DIFF
--- a/src/popup/components/tool.jsx
+++ b/src/popup/components/tool.jsx
@@ -1,8 +1,8 @@
 import PropTypes from 'prop-types';
-import React, { useContext } from 'react';
+import React from 'react';
 import { Link } from 'react-router-dom';
+import browser from 'webextension-polyfill';
 
-import EnvContext from '../env-context';
 import ErrorShape from '../utils/error-shape';
 import TicketShape from '../utils/ticket-shape';
 import Content from './content';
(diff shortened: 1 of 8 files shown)


src/popup/components/tool.jsx
```diff
@@ -1,18 +1,24 @@
...
 import NoTickets from './no-tickets';
 import TicketControls from './ticket-controls';
 
-function Tool({ tickets, errors }) {
-  const { close, openopts } = useContext(EnvContext);
+function close() {
+  window.close();
+}
 
+function openopts() {
```
klappradla, line 17: Did you on purpose decide against `openOpts`? Fine, it just took me a few seconds to parse it for the first time 😹

src/popup/render.jsx
```diff
@@ -15,16 +14,14 @@ function propped(Component, defaults) {
   return ProppedComponent;
 }
 
-function render(tickets, errors, env) {
+function render(tickets, errors) {
   const root = document.getElementById('popup-root');
 
   const element = (
-    <EnvContext.Provider value={env}>
```
klappradla, line 22: 🙏

klappradla (approved):
Awesome 😻 

Very good catch!

src/popup/components/tool.jsx
pmeinhardt: I only moved it without (re-)considering the name. Good catch, I'll improve the naming. 💚

PR #300, opened 2021-05-07
Try formatter package with build output removed

pmeinhardt:
Try using the formatter package from GitHub without included build:
https://github.com/bitcrowd/tickety-tick-formatter/pull/8

This PR is just here to try and see whether this works.
We can scrap this PR (or update it to reference the default formatter branch again).

I had to `rm -rf node_modules` before running `npm install` or `yarn install` again for the package to be built.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,13 +28,13 @@ commands:
     description: Restore dependencies
     steps:
       - restore_cache:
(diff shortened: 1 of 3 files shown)


pmeinhardt:
😞 Okay, this seems to not work properly yet. We should probably just go and publish proper releases for the formatter package.

PR #317, opened 2021-09-20
Update Jest

klappradla:
This updates Jest (and babel-jest) to patch a security vulnerability on a package the old versions depended on: `set-value`.

See: https://github.com/advisories/GHSA-4jqc-8m5r-9rpr

With the update, we're now on Jest v27. On that version, Jest uses the `node` environment as the default environment for running tests in, instead of the slower `jsdom` environment. This means, tests which use the DOM API need to be switched back to run in the `jsdom` environment.

See https://jestjs.io/blog/2021/05/25/jest-27#flipping-defaults for details about the change.

For projects where only some tests require the DOM API, the Jest team recommends to still run the whole suite in the `node` environment (faster) and recommend the `jsdom` environment only for those tests where it's actually needed.


pmeinhardt (approved):
Out of curiosity: Did you notice a difference in the time it takes to run the tests?

klappradla:
> Out of curiosity: Did you notice a difference in the time it takes to run the tests?

Haha, I was curious as well 🧐 - tried out two different setups:

1. Run in the `node` environment and only switch to `jsdom` where necessary (so essentially how it's set up in this PR)

2. Run everything in `json`

  Configured like this globally in the `jest.config.js` file: `testEnvironment: 'jsdom'`

Results: pretty much no difference 😉 

1. Multi environment:

    ```
    $ time yarn test
    …
    Time:        5.691 s, estimated 6 s
    Ran all test suites.
    ✨  Done in 6.64s.
    yarn test  27.86s user 6.21s system 493% cpu 6.909 total
    ```

2. `jsdom` environment

    ```
    $ time yarn test
    …
    Time:        6.226 s
    Ran all test suites.
    ✨  Done in 7.14s.
    yarn test  31.29s user 6.93s system 515% cpu 7.413 total
    ```

🤷

PR #323, opened 2021-09-30
Add Jest watch plugins for better DX

pmeinhardt:
~Based on #168.~

image

– https://github.com/jest-community/jest-watch-typeahead


klappradla (approved):
DX++

Great 👍 The way I understood it previously, was that this is a "feature" of newer versions of Jest and we'd get it via the upgrades anyway 🙈 

So plugin - also fine! Thanks!

PR #326, opened 2021-10-19
[#291] Convert Jira markup to markdown

### LINKED ISSUE #291: Convert Jira markup to markdown

pmeinhardt:
This needs a bit more testing. We should try a development build with Jira tickets we have in the wild.

The Jira API is queried in a different way. We should possibly check what this means in terms of compatibility, esp. with regard to on-premise Jira installations (which could be older?).

Closes #291.


pmeinhardt:
There's already a good improvement on the output.

Some weirdness I saw when trying it on tickets though:

- Messages include `&#x20;` in different places (which is just a "space" as far as I can tell)
- Escaped `:` on a URL from an `inlineCard` ADF node:`https\://bitcrowd.atlassian.net/browse`
- Backslashes in different places:
    - on line endings generated from an ADF `hardBreak` node
    - before dots in `text` nodes, e.g. `2\. double check…`
- Maybe we can improve rendering of `media` elements which currently output `<!-- media: file 80f76786-… -->`. I wonder whether we can derive their URL somehow

klappradla:
**Testing** 🥼 

- auto-format for the description (wrapping at 80 chars) does not work ❌

_(Just collecting some of my findings with the internal build here)_

klappradla:
Quick rebase 🏓

klappradla:
Just to confirm: played around a bit with the auto-formatting of the description and it - of course - works like charm 💯 

_(It also shouldn't really be affected by this change right here)_

Will do another round of test driving and then probably merge 🚢

klappradla (approved):
Pure gold ⛰️ 

The magic lies in the external packages (💪). Functionality is tested out in the real world, so we're good to go with that.

PR #329, opened 2021-11-17
Remove stacktrace-js

pmeinhardt:
We've not seen a huge benefit of having [stacktrace-js](https://www.npmjs.com/package/stacktrace-js) in the codebase so far and it hasn't been updated in a long time. 👋 

Removing it allows us to trim down on code and avoid additional complexity in error handling in the extension. ✂️ 

It does mean stacktraces posted as issues will possibly be more cryptic. However I can't remember the last time we received an issue report where we used the stacktrace to solve the problem. 🤷 

If we need to find the source location for a stack frame from a minified build, we should still be able to grab the source map for the given version and use something like [mozilla/source-map](https://github.com/mozilla/source-map) or [@yelo/source-map-cli](https://github.com/imyelo/source-map-cli) to retrieve the raw source location.


package.json
```diff
@@ -101,8 +101,8 @@
     "react-router": "5.2.1",
     "react-router-dom": "5.3.0",
     "react-textarea-autosize": "^8.3.2",
+    "serialize-error": "^9.0.0",
```
klappradla, line 104: Of course this person has a package for it 🦄

jest.config.js
```diff
@@ -14,7 +14,7 @@ module.exports = {
     "(?!^.+\\.(js|jsx|ts|tsx)$)": "<rootDir>/test/transforms/file.js",
   },
   transformIgnorePatterns: [
-    "node_modules/(?!copy-text-to-clipboard|ky|strip-indent)",
+    "node_modules/(?!copy-text-to-clipboard|ky|serialize-error|strip-indent)",
```
klappradla, line 17: ✨

src/popup/components/error-details.test.tsx
```diff
@@ -1,56 +1,15 @@
...
 
-jest
-  .mock("stacktrace-js", () => ({ fromError: jest.fn() }))
-  .mock("./copy-button", () => ({ value }: CopyButtonProps) => (
-    <button type="button" value={value}>
-      Copy error details
-    </button>
-  ));
-
-class MockStackFrame {
```
klappradla, line 17: 🧹

klappradla (approved):
Awesome idea 🎉 

Good point as well - I cannot remember anyone ever opened an issue where the copied stracktrace was important or of any help. Way cleaner with the approach you implemented ❤️

PR #330, opened 2021-11-30
Update type definitions

pmeinhardt:
A quick drive-by:

- Use ".ts" rather than ".d.ts" extension
- Update type documentation comments


src/additional.d.ts
```diff
@@ -1,5 +1,6 @@
 /// <reference types="react" />
 
+// Define type for ".svg" imports handled by "@svgr/webpack"
```
klappradla, line 3: 🙏

klappradla (approved):
Grande 🧀 

So, reading a bit through the Typescript docs, it seems like the use case for the `d.ts` would only be for declarations (or typing a JS package) 💡 

TIL

klappradla:
Quick rebase 🏓

PR #332, opened 2021-12-02
Update Safari build

pmeinhardt:
Remove content-hash from `icon.svg` so the XCode project reference stays constant.


anagromataf:
LGTM

PR #333, opened 2021-12-03
Explcitly set the language for the exported cert to en

anagromataf:
When using the script `script/extract-cert-ou` with another locale than en (e.g., de), the owner of the certificate can not be found ("Aussteller" vs. "Owner").


pmeinhardt (approved):
Neat! Thanks @anagromataf 🏳️

PR #336, opened 2022-03-06
Add Linear.app adapter

mtarnovan:
Adds a simple, DOM based [Linear](https://linear.app/) adapter.

The selectors are fragile (because Linear minimizes their CSS so we can't use their CSS classes) and will require updating if Linear adds more languages.

I had a few issues locally:
1. after running any tasks that create `dist/` the path is picked up by `yarn test` which produces a lot of confusing test failures
1. (_this is only an issue if one doesn't want to install Rosetta_) the build system uses some notifiers that come with pre-built x86 executables for `terminal-notifier`. Those will crash on arm64 (Apple M1). Since `terminal-notifier` seems unmaintained, as a (very) dirty work-around one can compile `terminal-notifier` to get an arm64 executable (for example via `brew`) and use that binary instead:
```
brew install terminal-notifier
ln -sf `which terminal-notifier` node_modules/node-notifier/node_modules/node-notifier/vendor/mac.noindex/terminal-notifier.app/Contents/MacOS/terminal-notifier
ln -sf `which terminal-notifier` node_modules/webpack-build-notifier/node_modules/node-notifier/vendor/mac.noindex/terminal-notifier.app/Contents/MacOS/terminal-notifier
```
In lieu of a better fix, maybe we should add this to the README?

### DIFF
--- a/src/core/adapters/linear.test.ts
+++ b/src/core/adapters/linear.test.ts
@@ -1,0 +1,43 @@
+/**
+ * @jest-environment node
+ */
+
+import { JSDOM } from "jsdom";
+
+import scan from "./linear";
+
+const pages = {
+  issuePage: {
+    url: new URL(
+      "https://linear.app/my_team/issue/FOO-1/do-something-interesting"
+    ),
+    document: "<span title='Edit issue title'>Do something interesting</span>",
+  },
+};
+
+describe("linear adapter", () => {
+  function doc(body = "") {
+    const { window } = new JSDOM(`<html><body>${body}</body></html>`);
+    return window.document;
+  }
+
+  it("returns an empty array if it is on a different page", async () => {
+    const result = await scan(new URL("https://example.com"), doc());
+    expect(result).toEqual([]);
+  });
+
+  it("extracts the ticket from the issue page", async () => {
+    const result = await scan(
+      pages.issuePage.url,
+      doc(pages.issuePage.document)
+    );
+    expect(result).toEqual([
+      {
+        id: "FOO-1",
+        title: "Do something interesting",
+        type: "issue",
+        url: pages.issuePage.url.toString(),
+      },
+    ]);
+  });
+});
--- a/src/core/adapters/linear.ts
+++ b/src/core/adapters/linear.ts
@@ -1,0 +1,34 @@
+/**
+ * Linear adapter
+ *
+ * The adapter extracts ticket information from the markup. Note that linear
+ * obfuscates classes, so the DOM selectors are probably not reliable (for one,
+ * they are locale dependent, but at this time Linear is only available in
+ * English).
+ *
+ */
+
+import { match } from "micro-match";
+
+import type { TicketData } from "../types";
+import { $text } from "./dom-helpers";
+
+async function scan(url: URL, document: Document): Promise<TicketData[]> {
+  if (url.hostname !== "linear.app") return [];
+
+  const { id, issueType } = match("/:team/:issueType/:id/:slug", url.pathname);
+  const title = $text('[title="Edit issue title"]', document);
+
+  if (!id) return [];
+
+  const ticket = {
+    id,
+    url: url.toString(),
+    title,
+    type: issueType,
+  } as TicketData;
+
+  return [ticket];
+}
+
+export default scan;
--- a/src/core/search.ts
+++ b/src/core/search.ts
@@ -4,6 +4,7 @@ import type { Ticket } from "../types";
 import GitHub from "./adapters/github";
 import GitLab from "./adapters/gitlab";
 import Jira from "./adapters/jira";
+import Linear from "./adapters/linear";
 import Notion from "./adapters/notion";
 import Tara from "./adapters/tara";
 import Trello from "./adapters/trello";
@@ -36,7 +37,15 @@ async function search(adapters: Adapter[], url: URL, document: Document) {
   return aggregate(results);
 }
 
-const stdadapters: Adapter[] = [GitHub, GitLab, Jira, Notion, Tara, Trello];
+const stdadapters: Adapter[] = [
+  GitHub,
+  GitLab,
+  Jira,
+  Notion,
+  Tara,
+  Trello,
+  Linear,
+];
 
 export { search, stdadapters };
 export default search.bind(null, stdadapters);
--- a/src/core/search.ts
+++ b/src/core/search.ts
@@ -41,10 +41,10 @@ const stdadapters: Adapter[] = [
   GitHub,
   GitLab,
   Jira,
+  Linear,
   Notion,
   Tara,
   Trello,
-  Linear,
 ];
 
 export { search, stdadapters };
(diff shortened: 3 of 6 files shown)


mtarnovan:
Note that we could also use Linear's SDK to fetch ticket info, but I gave it a try and it felt slow, so I think it's worth the tradeoff of using selectors even if might need updating from time to time.

.gitignore
```diff
@@ -17,3 +17,6 @@ dist/
 
 # Test coverage reports
 coverage/
+
+# vscode artifacts
+.vscode/
```
klappradla, line 22: How does this relate to the changes in this PR? I never used VSCode... does it generate these files automatically for every project? Or what are they for?

src/core/adapters/linear.test.ts
```diff
@@ -0,0 +1,43 @@
...
+    const { window } = new JSDOM(`<html><body>${body}</body></html>`);
+    return window.document;
+  }
+
+  it("returns an empty array if it is on a different page", async () => {
+    const result = await scan(new URL("https://example.com"), doc());
+    expect(result).toEqual([]);
+  });
+
+  it("extracts the ticket from the issue page", async () => {
```
klappradla, line 29: Are there also other views where one would highlight a ticket? E.g. a sprint or list view?

src/core/adapters/linear.ts
```diff
@@ -0,0 +1,34 @@
...
+
+async function scan(url: URL, document: Document): Promise<TicketData[]> {
+  if (url.hostname !== "linear.app") return [];
+
+  const { id, issueType } = match("/:team/:issueType/:id/:slug", url.pathname);
+  const title = $text('[title="Edit issue title"]', document);
+
+  if (!id) return [];
+
+  const ticket = {
```
klappradla, line 24: Is there also a `description` field in a somewhat commit-message-body-usable format?

src/core/search.ts
```diff
@@ -36,7 +37,15 @@ async function search(adapters: Adapter[], url: URL, document: Document) {
...
 
-const stdadapters: Adapter[] = [GitHub, GitLab, Jira, Notion, Tara, Trello];
+const stdadapters: Adapter[] = [
+  GitHub,
+  GitLab,
+  Jira,
+  Notion,
+  Tara,
+  Trello,
+  Linear,
```
klappradla, line 47: Let's stick with alphabetic ordering 🤷 Even though it's not "fair".

src/core/adapters/linear.ts
```diff
@@ -0,0 +1,34 @@
...
+import { match } from "micro-match";
+
+import type { TicketData } from "../types";
+import { $text } from "./dom-helpers";
+
+async function scan(url: URL, document: Document): Promise<TicketData[]> {
+  if (url.hostname !== "linear.app") return [];
+
+  const { id, issueType } = match("/:team/:issueType/:id/:slug", url.pathname);
+  const title = $text('[title="Edit issue title"]', document);
```
klappradla, line 20: This is indeed a bit fragile 🧱

klappradla (commented):
Hey @mtarnovan 👋 

Thanks a ton for your interest in this project and for jumping on board with a contribution 💚 

I'm a bit hesitant about integrating this adapter though: I've never heard of the tracker software and we're not using it internally. This combined with the fact that the adapter's implementation is - as you said - very fragile and likely to break on any updates the team does to their software (imagine for instance they switch from `<span>` with `title` attribute to normal `<label>` and `<input>` tags), makes it quite hard to maintain for us.

And having an adapter which is very likely to break fully unnoticed is a bit of a pain to maintain for us.

I'll think about it again and hopefully have the time some day this week to also try it with the software.

Thanks a ton 💙

.gitignore
mtarnovan: It doesn't directly relate to the changes. This dir is where VSCode keeps some per project files (such as build task definitions etc)

src/core/adapters/linear.ts
mtarnovan: The markup looks like this:

Screenshot 2022-03-07 at 17 01 01

So I don't see a way around this other that the one I already mentioned: use the official SDK which is a client for a Graphql API (but as I said I played around with it seems a bit slow, I don't want to wait seconds to grab a branch command 🙃). Ofc using an API client would be much more reliable and maintainable, so if that's your preference I could explore that alternative.

src/core/adapters/linear.ts
mtarnovan: If we use the API (see my [other comment](https://github.com/bitcrowd/tickety-tick/pull/336#discussion_r820795967)) then yes, otherwise, probably not very reliably.

mtarnovan:
Hey Max 👋 

> I thought about this PR again over the holiday. It would indeed be great if TicketyTick would support a more diverse set of trackers, even tools which bitcrowd isn't using internally. If you and your team are actively using the tool, we can be sure that breaking changes on the Linear.app side are detected quite fast.

We are probably going to stick with Linear for the foreseeable future, so yes, we can be the canary in the coal mine :-)
I actually replaced the release extension with a dev build from this PR and have been using that at work in the last few days (we don't use the description field in commits anyway since it feels redundant)

> So my suggestion would be: do a quick spike on a) trying to get the ticket's description field somehow and b) base the fetching on the API rather than the DOM. If both don't work out, I'd also be fine with merging the current approach from this PR.

Happy to give that a try when time permits.

> Another note on fetching via the API: take a look at the notion.so and Jira adapters: it can make sense to use the internal API the SPAs are using, rather than a public API the tools offer. When building the notion.so adapter, I just opened the network tab in the developer tools to see what request the app does. As the client is already authenticated, we can just do the same requests in the extension 🤷

From my research Linear uses the same GraphQL API internally and externally (they provide an SDK that wraps it). They provide a global function `loadLinearClient` that loads an already authenticated client. My first implementation was using that, but it felt slow so I threw it out (as I mentioned above for us the `description` field is not at all important, since Linear has a Github integration that already takes care of that for us by adding the description to the PR automatically; so I prefered keeping the code fast and simple by not accessing any APIs)

I'll try what you're suggesting, but I'm pretty sure it will not work: their internal APIs are authenticated with Oauth bearer tokens, not with session cookies (like I assume notion.so does).

To summarize:
- the Linear adapter will probably have the limitation of not working on "index" views (sprint, board, roadmaps etc) since the markup makes it very difficult to reliably iterate over individual tickets from those pages; for us this is not important at all, but ofc I can't speak for other users of tickety-tick
- I'll work on your suggestions to integrate the API and get the description field too

Thanks again,
Mihai

.gitignore
andreasknoepfle: I think this belongs into you [own global gitignore ](https://stackoverflow.com/questions/7335420/global-git-ignore) config, since it is quite specifc to your editor setup :). 

```suggestion
```

mtarnovan: Totally forgot about the global gitignore, thanks!

klappradla (approved):
Thanks for the updates @mtarnovan 🙏 

If this has been working for you with a local build in the last weeks, let's give this a try for everyone 🌊 ⛵

mtarnovan:
@klappradla nice! 

Please ping me when you release this to the extension stores so I can replace the locally build extension. I'll keep an eye on breakages and will try to fix them asap.

PR #342, opened 2022-04-25
Small fixes and updates

klappradla:
Some small cleanup fixes and updates.

Changes:

- Fix warning in tests

  We were nesting a `ul` element into a `p` element - which is invalid. As this was only within a mock, I just changed the mock to a plain `span` which renders the list inline.

- Update browserlist

- Upgrade web-ext package

  This should free the path to more dependency updates.

### DIFF
--- a/src/popup/components/__snapshots__/no-tickets.test.tsx.snap
+++ b/src/popup/components/__snapshots__/no-tickets.test.tsx.snap
@@ -18,13 +18,9 @@ exports[`no-tickets with errors renders an error report 1`] = `
       <br />
       while scanning for tickets.
     </p>
-    <p />
-    <ul>
-      <li>
-        Error: An error to test error reporting
-      </li>
-    </ul>
-    <p />
+    <p>
+      - Error: An error to test error reporting
+    </p>
     <p>
       <a
         href="https://github.com/bitcrowd/tickety-tick/issues"
--- a/src/popup/components/no-tickets.test.tsx
+++ b/src/popup/components/no-tickets.test.tsx
@@ -6,11 +6,7 @@ import type { Props } from "./no-tickets";
 import NoTickets from "./no-tickets";
 
 jest.mock("./error-details", () => ({ errors }: ErrorDetailsProps) => (
-  <ul>
-    {errors.map((error, index) => (
-      <li key={index}>{error.toString()}</li> // eslint-disable-line react/no-array-index-key
-    ))}
-  </ul>
+  <>{errors.map((error) => `- ${error}`).join("\n")}</>
 ));
 
 describe("no-tickets", () => {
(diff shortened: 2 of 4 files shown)


tessi (approved):
💪 💯

tessi (approved):

tessi:
ahh, I can't properly approve anymore. makes sense :D
thanks for keeping things updated here! 💛

src/popup/components/__snapshots__/no-tickets.test.tsx.snap
```diff
@@ -18,13 +18,9 @@ exports[`no-tickets with errors renders an error report 1`] = `
       <br />
       while scanning for tickets.
     </p>
-    <p />
```
pmeinhardt, line 21: This is a "self-closing" `p`-tag, it should be equivalent to `<p></p>`. While I can't tell where these are from based on the diff here, it doesn't look like the `ul` was nested inside a `p` like you mention in the PR description. 🤔

pmeinhardt: For reference: https://reactjs.org/docs/jsx-in-depth.html

pmeinhardt: Since it's only a mocked case it doesn't matter much. I just noticed and was wondering what error message you saw. 🙂

pmeinhardt: Ah. That makes sense. They're wrapped one level above. 👍

src/popup/components/no-tickets.test.tsx
```diff
@@ -6,11 +6,7 @@ import type { Props } from "./no-tickets";
 import NoTickets from "./no-tickets";
 
 jest.mock("./error-details", () => ({ errors }: ErrorDetailsProps) => (
-  <ul>
-    {errors.map((error, index) => (
-      <li key={index}>{error.toString()}</li> // eslint-disable-line react/no-array-index-key
-    ))}
-  </ul>
+  <>{errors.map((error) => error.toString()).join("\r\n")}</>
```
pmeinhardt, line 9: Is there a reason to use `"\r\n"` rather than just `"\n"`? Not criticizing Just curious. 🙂 

If you want to emphasize the fact that the passed errors are a list, I could imagine:

```suggestion
  <>{errors.map((error) => `- {error}`).join("\n")}</>
```

could highlight this a bit more. It doesn't have any practical impact though so no need to change.

pmeinhardt (approved):
Thanks for maintaining this project @klappradla 💖

src/popup/components/no-tickets.test.tsx
klappradla: Hm, now that you point towards it, I actually have no clue what the `\r` does 🤷 I just quickly searched for the best and most secure way to do line breaks in `JS`. But after all, I haven't seen `\n` fail, so let's drop the `\r`.

For the bullet point part: yes, I was already thinking about the same 💚  Then let's do this!

Issue #159, opened 2019-05-20
Add a screencast to the readme?


pmeinhardt:
Should we add a short screencast to the README to demo how Tickety-Tick works and give people coming across this repo a better idea about it?

klappradla:
💥 awesome advertising idea!

Let's schedule this for our next hacknight / lunchhacking.

klappradla:
Another person requested this: https://github.com/syntax-tree/mdast/pull/40#issuecomment-1047578714 📌

klappradla:
Plan of attack: run a script to show the core features.

Record using `kap.app` (it has a plugin to show keys pressed). We may need to edit the result to remove gaps.

klappradla:
Made a tiny attempt 🍿 

![tt-demo-small](https://user-images.githubusercontent.com/3491815/165542748-badc6687-2255-4a98-9d0e-458e8043fbbd.gif)

PR #344, opened 2022-04-28
[#159] Add a screencast to the readme

klappradla:
#159

Update README with a short screencast of a possible workflow using Tickety-Tick 🍿


ammancilla (approved):
🎥

planktonic (approved):
🪄

klappradla:
Oh, that's great news @pmeinhardt 👀 
I just tried `.webm` - which isn't supported... Will have a quick look into how an `mp4` or `mov` turns out.

klappradla:
Ok, took me a minute to figure out @pmeinhardt: embedding videos only works when uploaded via drag & drop 😂 

Fair enough 🤷 Updated with an mp4 version.

PR #351, opened 2022-10-11
Update CI base image

klappradla:
Use the faster next gen image for NodeJS.

See [this blog post](https://circleci.com/blog/announcing-our-next-generation-convenience-images-smaller-faster-more-deterministic/) for a reference.

Unfortunately, for the Swift image we're using, Circle does not have their own images.


iampeternguyen (approved):
🚢

Issue #349, opened 2022-10-07
No longer working with Jira Server


tiagotex:
The newly released version updated the Jira adapter to depend on API `3` instead of `latest`. https://github.com/bitcrowd/tickety-tick/pull/326/
 
This is a problem because API v3 is only available for Jira Cloud, Jira Server only has v2.

https://community.atlassian.com/t5/Jira-questions/When-will-Jira-Server-support-REST-API-v3/qaq-p/1303614
https://jira.atlassian.com/browse/JRASERVER-70688

Would you be open to reverting this functionality and reworking the feature to maintain compatibility with Jira Server?

Of course, I'm available to make the necessary changes mentioned above.

Thank you for the awesome plugin 💚 

PS: this might be the same problem affecting @RKushnir  https://github.com/bitcrowd/tickety-tick/issues/345

RKushnir:
Hey Tiago 🤗 I no longer need to use that Jira instance, so it doesn't bother me anymore. But I hope you can figure it out.

tiagotex:
Created some instructions to use previous version of the extension on Chrome while this is addressed:

```
1. Go to https://github.com/bitcrowd/tickety-tick/releases/tag/v4.2.3 and download the chrome version
2. Navigate to chrome://extensions in Chrome
3. Remove the current tickty-tick extension by clicking "Remove"
4. Expand the Developer dropdown menu and click “Load Unpacked Extension”
5. Navigate to the local folder containing the extension’s code and click Ok
```

klappradla:
Hey @tiagotex,

thanks for reporting 💚 

From what I remember, Atlassian only from API v3 on supports serving ticket descriptions in their proprietary ADF format (rich text document format). And our updated JIra adapter relies on that format. So simply switching it back to `/latest` probably won't be enough.

We need a moment to think about a good way to support both systems (unfortunately I don't have access to a Jira server instance...). If you have time and could do me a favor, it would be nice if you could post me an example response for a request like that:

```
https://my-subdomain.atlassian.net/jira/software/projects/TT/boards/8?selectedIssue=…}
```

I am particularly interested in how the `description` field looks like when you use links, tables, etc. in the ticket description.

klappradla:
[...]

PR #354, opened 2022-10-20
[#349] No longer working with Jira Server

andreasknoepfle:
https://github.com/bitcrowd/tickety-tick/issues/349


src/core/adapters/jira-cloud.ts
```diff
@@ -1,5 +1,5 @@
 /**
- * Jira adapter
+ * Jira Cloud adapter
```
maltoe, line 2: Maybe a word of what the difference between Cloud and Server is?

src/core/adapters/jira-server.test.ts
```diff
@@ -0,0 +1,177 @@
...
+
+const ticket = {
+  id: response.key,
+  title: response.fields.summary,
+  description: response.fields.description,
+  type: response.fields.issuetype.name.toLowerCase(),
+  url: `https://my-domain.com/browse/${key}`,
+};
+
+describe("jira adapter", () => {
```
maltoe, line 32: jira server

src/core/adapters/jira-server.test.ts
```diff
@@ -0,0 +1,177 @@
...
+    expect(client).toHaveBeenNthCalledWith(2, endpoint);
+    expect(client).toHaveBeenNthCalledWith(3, endpoint);
+
+    expect(results).toEqual([
+      [expectedTicket],
+      [expectedTicket],
+      [expectedTicket],
+    ]);
+  });
+});
```
maltoe, line 177: you could probably deduplicate a bunch of code here... but maybe now :shrug:

src/core/search.ts
```diff
@@ -40,7 +41,8 @@ async function search(adapters: Adapter[], url: URL, document: Document) {
 const stdadapters: Adapter[] = [
   GitHub,
   GitLab,
-  Jira,
+  JiraCloud,
+  JiraServer,
```
maltoe, line 45: test for this? That JiraCloud takes precendence over JiraServer?

maltoe (approved):
hooray :tada:

src/core/adapters/jira-server.ts
```diff
@@ -0,0 +1,90 @@
+/**
+ * Jira Server adapter
+ *
+ * This adapter is for self-hosted jira instances and uses the old rest API for fetching tickets.
```
klappradla, line 4:
```suggestion
 * This adapter is for self-hosted jira instances and uses the old v2 rest API for fetching tickets.
```

klappradla (commented):
Awesome work @andreasknoepfle 💯 

Did you get the chance to test this on both versions of Jira, one with the v3 API and one with only the v2 API?

klappradla:
Just tested on Jira Cloud with a few tickets - still the same as before 💚  nice!

andreasknoepfle:
> Did you get the chance to test this on both versions of Jira, one with the v3 API and one with only the v2 API?

yes, tested both and yesterday both worked nicely (before the review changes though... we might want to test again)

klappradla (approved):
Looking good ✨

Differentiating which adapter to use via URL and ID on the body is way simpler than the initial approach I had in mind 😮 

Tested it again locally and it seems to work just fine. I wasn't 100% sure whether the URL for Jira cloud was always `.net`, as Atlassian itself has `.com`. But for now I could not find any other information that people could configure that. So let's roll with this for now.

src/core/search.test.ts
```diff
@@ -45,4 +47,10 @@ describe("ticket search", () => {
       errors: [error0, error1].map(serialize),
     });
   });
+
+  it("searches jira cloud before jira server", () => {
+    expect(stdadapters.indexOf(JiraCloud)).toBeLessThan(
+      stdadapters.indexOf(JiraServer)
+    );
+  });
```
pmeinhardt, line 55: I have a question and a remark:

1. Why does the order matter here?
2. Even though the adapters are listed in this order, it doesn't mean the search with the cloud adapter will wait for the first to complete before trying the next one. If you look at `src/core/search.ts` (`search` function) you'll see that we `map` across all adapters and then `await` with `Promise.all` – meaning the adapter scans are run concurrently.

pmeinhardt: If you want to guarantee the cloud adapter scan completes before attempting the Jira Server scan, you need to either change the `search` behavior` or do it within one adapter which is why I proposed this in the issue:
https://github.com/bitcrowd/tickety-tick/issues/349#issuecomment-1274536884

If you're sure they'll never both complete successfully for any given page, you don't need to care about the order. Assuming one succeeds, but the other fails, errors will be discarded and the successful result used. Ideally though, you should fail as early as possible and avoid unnecessary network requests if you can identify which kind of Jira you're on.

klappradla: Thanks for jumping onto this (again) @pmeinhardt 💙 

From a first glance, you're of course absolutely right: nothing keeps the `jira-server` adapater to be ready with results **before** the `jira-cloud` one on cloud instances. The `jira` ID on the body tag is there for both types of installations.

It's probably more future-proof to handle that in one adapter. This PR should for now have fixed "not working on Jira server" for now. I'll open an issue for a cleanup to merge the two adapters. I'd personally prefer it over changing the behavior of the `search` function.

See https://github.com/bitcrowd/tickety-tick/issues/359

Issue #362, opened 2022-11-15
expose notion page url 


bitboxer:
The notion ticket id is a sad guid. I would love to be able to write a template that creates a markdown link to the ticket. Sadly the url is not available in the template.

klappradla:
Nice to know someone is using the notion adapter 🤗 

I don't remember why I did not include the URL back then - maybe the templates weren't as sophisticated as they're now and we did not update the adapter to fit it.

Feel free to open a PR with the update if you need it right away @bitboxer. It will take a bit of time until I come around to do it myself. But it's definitely something to be done 👍

PR #368, opened 2023-01-11
[#362] Expose notion page url

jwharrow:
This PR adds the URL to a ticket in the Notion adapter. It will work on both the ticket show and the board view (modal). It relies on Notion redirecting from `/id` to `/slug`.

https://github.com/bitcrowd/tickety-tick/issues/362

### DIFF
--- a/src/core/adapters/notion.ts
+++ b/src/core/adapters/notion.ts
@@ -53,7 +53,7 @@ function getSelectedPageId(url: URL) {
   const slugId = isPageModal ? params.get("p") : getPageFromPath(path);
 
   if (!slugId) return null;
-  return uuid(slugId);
+  return slugId;
 }
 
 function extractTicketInfo(result: NotionTicketInfo) {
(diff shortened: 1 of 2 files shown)


src/core/adapters/notion.ts
```diff
@@ -86,7 +92,10 @@ async function scan(url: URL): Promise<TicketData[]> {
     .post("api/v3/getRecordValues", request)
     .json<NotionTicketResponse>();
 
-  return getTickets(response, id);
+  return getTickets(response, id).map((ticket) => ({
+    url: buildUrl(url, slugId),
```
klappradla, line 96: As you mentioned yesterday: How about taking the URL and only setting the path to the `slugId`. So we end up with something like `https://www.notion.so/8c2499ce59654658a1037c11bdd50523` which will be redirected accordingly by notion itself.

Does this fix the problem for non-organization users? I forgot the details about it, sorry 🙈

klappradla (commented):
Awesome team work @hannahbit & @jwharrow 👯 

Looks perfect ✨ Just wondering if we could fix the situation for non-orga users...

src/core/adapters/notion.ts
hannahbit: > Does this fix the problem for non-organization users? I forgot the details about it, sorry 🙈

Yes! Will add a commit to change this.

klappradla (approved):
Thanks @hannahbit 🦄

Issue #369, opened 2023-01-12
Support notion pages without organizations


jwharrow:
Currently a URL like `notion.so/:slug" is not supported -- "notion.so/:organization/:slug" is what is expected.

PR #370, opened 2023-01-12
[#369] Support notion pages without organizations

jwharrow:
Removes dependence on :organization in ticket path
https://github.com/bitcrowd/tickety-tick/issues/369

### DIFF
--- a/src/core/adapters/notion.ts
+++ b/src/core/adapters/notion.ts
@@ -9,6 +9,8 @@
  * API yet.
  */
 
+import { match } from "micro-match";
+
 import client from "../client";
 import type { TicketData } from "../types";
 
@@ -41,7 +43,7 @@ function uuid(slugId: string | null) {
 }
 
 function getPageFromPath(path: string) {
-  const slug = path.split("/").slice(-1)[0];
+  const { slug } = match(":organization?/:slug", path);
   if (!slug) return null;
 
   return slug.replace(/.*-/, ""); // strip title from slug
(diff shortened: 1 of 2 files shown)


hannahbit (approved):
LGTM 🌈 

Thanks!

src/core/adapters/notion.ts
```diff
@@ -41,7 +41,7 @@ function uuid(slugId: string) {
 }
 
 function getPageFromPath(path: string) {
-  const { slug } = match("/:organization/:slug", path);
+  const slug = path.split("/").slice(-1)[0];
```
klappradla, line 44: Let's keep using the `micro-match` package here:

```js
const { slug } = match(":organization?/:slug", path);
```

It can do optional parameters.

src/core/adapters/notion.ts
```diff
@@ -30,7 +28,9 @@ type NotionTicketResponse = {
  * Turns a slugged page ID without dashes into a dasherized RFC 4122 UUID.
  * UUID-Format: 96ec637d-e4b0-4a5e-acf3-8d4d9a1b2e4b
  */
-function uuid(slugId: string) {
+function uuid(slugId: string | null) {
+  if (!slugId) return null;
```
klappradla, line 32: Hm, I to be honest don't get the point of this change 🤔. Maybe I'm missing something... But why would we want to spread checks for the ID being `null` over multiple functions? How about just an early return inside the `scan` function before even calling `uuid`?

klappradla (changes requested):
Neat! I have some minor comments, will quickly add a commit to this PR myself 🚜

klappradla:
@jwharrow || @hannahbit can you check the two commits I added?

iampeternguyen (approved):
changes looks good 🚢

klappradla (approved):

klappradla:
Nice 🎉 We should schedule a release then soon 🚢

PR #379, opened 2023-03-17
Add adapter for YouTrack

pmeinhardt:
Heya! 👋 I hope you're all doing great.

Here's a new adapter for [YouTrack](https://www.jetbrains.com/youtrack/).

_Disclaimer: I haven't tested this on a public cloud instance, configuration and markup might look slightly different there. But I guess we can improve support in case issues about this adapter do come in._

### DIFF
--- a/src/core/adapters/youtrack.test.ts
+++ b/src/core/adapters/youtrack.test.ts
@@ -1,0 +1,63 @@
+/**
+ * @jest-environment node
+ */
+
+import { JSDOM } from "jsdom";
+
+import scan from "./youtrack";
+
+const html = {
+  issuepage: (type?: string) => `
+    <div class="yt-issue-view">
+      <span class="js-issue-id">TT-123</span>
+
+      <h1 class="yt-issue-body__summary" data-test="issueSummary">
+        Support YouTrack
+      </h1>
+
+      <button type="button" data-test="Type">
+        <span>${type ?? ""}</span>
+      </button>
+    </div>
+  `,
+};
+
+const url = new URL(
+  "https://bitcrowd.youtrack.cloud/issue/TT-0/Support-YouTrack"
+);
+
+describe("youtrack adapter", () => {
+  function doc(body = "") {
+    const { window } = new JSDOM(`<html><body>${body}</body></html>`);
+    return window.document;
+  }
+
+  it("returns an empty array if it is on a different page", async () => {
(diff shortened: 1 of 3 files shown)


src/core/adapters/youtrack.ts
```diff
@@ -0,0 +1,24 @@
...
+import { $has, $text } from "./dom-helpers";
+import { hasRequiredDetails } from "./utils";
+
+async function scan(url: URL, document: Document): Promise<TicketData[]> {
+  if (!$has(".yt-issue-view", document)) return [];
+
+  const id = $text(".js-issue-id", document)?.trim();
+  const title = $text("[data-test='issueSummary']", document);
+  const type =
+    $text("[data-test='Type']", document)?.toLowerCase().trim() || "feature";
```
klappradla, line 17: Weird that the word "type" is capitalized here. But you'll probably notice quite fast when then changes...

src/core/adapters/youtrack.test.ts
```diff
@@ -0,0 +1,63 @@
+/**
+ * @jest-environment node
+ */
+
+import { JSDOM } from "jsdom";
+
+import scan from "./youtrack";
+
+const html = {
+  issuepage: (type?: string) => `
```
klappradla, line 10: Is this structure supposed to leave room for quickly adding support for other pages in the future? Sweet 🍬 

From the product page it looks like there's a Jira-like backlog view where one quick-views issues. That could could probably be a candidate 👀

klappradla (approved):
Nice one 🍭 

Always in for a good old DOM-adapter 🙌. Didn't know this YT-tool existed 💡

Solid implementation as always 🪨. I briefly wondered about the early return in `scan`, because the CSS selector looks generic enough to be present on other tracker pages as well. But with checking the URL not being an option and scanning with this adapter being pretty fast, I'd just roll with it like it is.

src/core/adapters/youtrack.test.ts
pmeinhardt: Yep, exactly. I took inspiration from the existing Jira adapter when I whipped up this PR. It seemed like there might be other pages to extract tickets from and/or subtle differences between the internal JetBrains YouTrack instance and cloud-hosted instances. I can imagine that there's on-premise instances out there too… To get started I decided to just go for this one case.

Issue #372, opened 2023-01-19
Additional information for the JIRA adapters


andreasknoepfle:
For us it would be super useful if we could have the jira EPIC and software component as a configurable placeholder, that does not do anything for the other adapters. Like this it could very well be in the default set still.

andreasknoepfle:
Due to some other ideas in releasing, this just got a bit less of an issue. But I guess it would still be nice to have.

pmeinhardt:
Great idea for a new feature. 👌 

<br>

I think that would entail some more significant changes to the templating. Here's why I think so:

You **probably** want more logic in the templates in this case. Rather than not rendering anything in place of, say,

```
… {{ jira.epic }} …
```

 you'll probably want something more like this:

```
… {% if jira.epic !== '' %} {{ jira.epic }}{% endif %} …
```

Otherwise you'll end up with whitespace/formatting inconsistencies when trying to use the same template for Jira and non-Jira projects.

It's possible, but will make the template implementation more complex (and possibly more error-prone).

<br>

For some additional **background**:

In a previous iteration, when I first added the configurable message/branch/command templates to Tickety-Tick, I actually went for a proper templating system first. I don't remember which ones I've tried, but the issue all of those existing ones had, was that in the web-extension, certain JavaScript features are disallowed. Notably [`eval()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/eval) and [`Function("…")`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/Function) to dynamically evaluate template expressions raised exceptions.

I think more expressive templates (ideally based off an existing template syntax) would be a great addition to Tickety-Tick. It will probably take some re-evaluation of the template engines out there and checking them in a web-extension (security) context or - as pointed out - implementing at least conditional template sections.

<br>

Just my 2 cents. 🤷 

I may have overlooked a simpler but still user-friendly approach.


Happy to brainstorm some more or flesh out further steps. ❤️

klappradla:
As this topic is not as pressing as it was (because of Jira automations, Moco, etc.) I'd close this for now in order to keep TicketyTick as simple as it is.

Of course, feel free to reopen the discussion whenever you want or think it's necessary 🤗

PR #383, opened 2023-04-02
New template engine

pmeinhardt:
### Description

This integrates [LiquidJS](https://liquidjs.com/) for rendering branch, commit and command templates in Tickety-Tick. This adds the following features:

- Additional filters, see: https://liquidjs.com/filters/overview.html
- Conditionals, loops etc.: https://liquidjs.com/tags/overview.html

to use inside templates.

The syntax is similar to, but not the same as the template syntax previously used by Tickety-Tick.

Most importantly, LiquidJS does not use `eval()` or `new Function()`:

When released, we should publish an announcement and migration guide, possibly find a way of showing a notification in the browser/extension popup to notify users about the breaking changes and the need to update custom templates in the extension settings.

### State

This still needs further testing to verify it provides the desired functionality for all supported browsers. So far I've tested in Firefox which seems to work well.

TODO:

- [ ] Check browser-support more thoroughly:
  - [ ] Firefox
  - [ ] Chrome
  - [ ] …
- [ ] Update settings page
  - [ ] Simplify state handling, async preview updates etc.
  - [ ] Improve layout for the longer list of template filters at the bottom
  - [ ] Update tests
- [ ] Check whether we want to notify users about the changed template syntax in the extension popup and/or options page (for those who use customized templates)
- [ ] Create template migration guide?
  - [ ] `lowercase` => `downcase`
  - [ ] …
  - [ ] changed syntax for passing filter params

Relates to #372.


klappradla:
Some ideas 🧠 

* Link the filter names on the setting screen to their documentation (e.g https://liquidjs.com/filters/escape.html)
* For the breaking changes (break existing custom templates):
  * store Tickety-Tick version with customizations
  * if a person has a customization **without version**, show them a migration hint in the popup window (link to a `MIGRATION.md` in the repo or the release notes)

klappradla:
Hey @pmeinhardt 👋 

As this topic did not come up again for more than a year now, I'd vote to close this and stick to the templates as they are for now. I would want to avoid introducing more complexity if not absolutely necessary. And as it seems right now, only a fraction of users even makes use of the template customizations 😉 

Feel free to reopen whenever you want and be sure to have me on board for it 💚

PR #395, opened 2023-09-07
Update browserlist DB

klappradla:
This is maintenance we need to do now and then, to make sure our browser-compatibility is up to date.

See https://github.com/browserslist/update-db#readme


pmeinhardt (approved):

klappradla:
Superseeded by other stuff already...

PR #396, opened 2023-09-08
Update dependencies part 1

klappradla:
The dependencies in this project are pretty outdated. This is the start of a small series to get this back up to speed. The aim is to also make applying security patches easier through that.

**Changes in this PR:**

- Update browserlist DB
- Update babel packages
- Update webpack packages
- Update typescript
- Update prettier and related packages
- Fix code formatting issues

The most interesting part here is probably the update of `prettier` and untangling that we were requiring it as a dev dependency **and** a runtime dependency. The latter most likely came from reverting a commit which years ago extracted the formatter into a separate package.

As of version 3 of `prettier`, the default setting for trailing commas is "all", which boils down to: insert trailing commas whenever possible in any multi-line comma-separated list (see [their docs](https://prettier.io/docs/en/options.html#trailing-commas)). In modern JS, this also includes function calls (see [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas) for a reference). A motivation for this is probably cleaner diffs.

I ran `prettier` with the `fix` option to insert the missing commas.


src/core/format/pretty-print.ts
```diff
@@ -1,4 +1,4 @@
-import markdown from "prettier/parser-markdown";
+import markdown from "prettier/plugins/markdown";
 import prettier from "prettier/standalone";
```
pmeinhardt, line 2: Thanks for keeping things fresh @klappradla. 💚 

I only had a quick look, but it seems like you might need to update this import:

```suggestion
import * as prettier from "prettier/standalone";
```

🍋

klappradla: Yup 💜 

Already been there, but not quite dialed in yet. The interface also change to async...

andreasknoepfle (approved):
Lgtm now 🚀

PR #402, opened 2024-02-23
[401] Fix GitHub adapter bug label detection

klappradla:
This updates the GitHub adapter for the new Markup GitHub uses for labels on issues. With this change, issues labelled as "bug" should again result into a branch name like `bug/xyz/`.

On the side, this also removes the code for selecting issues on GitHub project pages. GitHub projects are entities on the user/organization now, not part of a repository. If we want to support them, it needs to be done differently. There's a follow up issue for that #403.

### DIFF
--- a/src/core/adapters/github.ts
+++ b/src/core/adapters/github.ts
@@ -18,9 +18,9 @@ import { hasRequiredDetails } from "./utils";
 export const selectors = {
   default: {
     issuesPage: ".js-check-all-container .js-issue-row.selected",
-    issuesPageLabel: ".labels .IssueLabel",
+    issuesPageLabel: ".lh-default .IssueLabel",
     issuePage: ".js-issues-results .gh-header-number",
-    issuePageLabel: '.js-issue-labels .sidebar-labels-style[title="bug"]',
+    issuePageLabel: '.js-issue-labels .IssueLabel[data-name="bug" i]',
   },
   legacy: {
     issuesPage: ".issues-listing .js-issue-row.selected",
--- a/src/core/adapters/github.ts
+++ b/src/core/adapters/github.ts
@@ -38,16 +38,21 @@ async function attempt(
   if ($has(select.issuesPage, doc)) {
     const issues = $all(select.issuesPage, doc);
 
-    const tickets = issues.map((issue) => {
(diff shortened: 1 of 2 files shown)


src/core/adapters/github.ts
```diff
@@ -18,9 +18,9 @@ import { hasRequiredDetails } from "./utils";
 export const selectors = {
   default: {
     issuesPage: ".js-check-all-container .js-issue-row.selected",
-    issuesPageLabel: ".labels .IssueLabel",
+    issuesPageLabel: ".lh-default .IssueLabel",
     issuePage: ".js-issues-results .gh-header-number",
-    issuePageLabel: '.js-issue-labels .sidebar-labels-style[title="bug"]',
+    issuePageLabel: '.js-issue-labels .IssueLabel[data-name="bug" i]',
```
klappradla, line 23: Note: the `i` is for selecting case-insensitive.

src/core/adapters/github.ts
```diff
@@ -38,16 +38,21 @@ async function attempt(
   if ($has(select.issuesPage, doc)) {
     const issues = $all(select.issuesPage, doc);
 
-    const tickets = issues.map((issue) => {
+    const tickets = issues.reduce((acc, issue) => {
       const id = $value("input.js-issues-list-check", issue);
       const title = $text("a.js-navigation-open", issue);
+
+      if (!id || !title) return acc;
```
klappradla, line 45: The Typescript compiler complained, that we cannot have `Null` titles in `TicketData`. That's actually correct and I don't know why it wasn't a problem before...

src/core/adapters/github.ts
```diff
@@ -38,16 +38,21 @@ async function attempt(
   if ($has(select.issuesPage, doc)) {
     const issues = $all(select.issuesPage, doc);
 
-    const tickets = issues.map((issue) => {
+    const tickets = issues.reduce((acc, issue) => {
```
klappradla, line 41: Turned this into a `reduce`, so that we can add the check for `id` and `title` to potentially "skip" an entry. Makes the compiler happy, but is unlikely to ever happen in the real world.

maltoe (approved):
Nice job! :100:

PR #405, opened 2024-04-04
Prepare v5.2.0

klappradla:
Bump version to v5.2.0.


klappradla:
> We do not have a CHANGELOG here, right?

Only in the "releases" section. It will be generated by the release script 🤓 _(let's see...)_

PR #407, opened 2024-04-16
[406] Fix rendering of preferences page

klappradla:


src/options/components/form.test.tsx
```diff
@@ -87,7 +89,9 @@ describe("form", () => {
 
     expect(checkbox).not.toBeChecked();
     expect(format).toHaveBeenLastCalledWith(expect.any(Object), false);
-    expect(screen.getByText("formatted-commit (false)")).toBeInTheDocument();
+    await waitFor(() => {
+      expect(screen.getByText("formatted-commit (false)")).toBeInTheDocument();
```
klappradla, line 93: 🥇  gold! I previously changed it to `findByText`, so that it waits - but no luck 😢

Issue #406, opened 2024-04-16
Preferences page broken


klappradla:
After switching to prettier's async API, the preference page can no longer be rendered. Thanks for reporting @Joshua-at-bitcrowd 🙏 

![image](https://github.com/bitcrowd/tickety-tick/assets/3491815/cf528d0b-a197-462a-9bc7-3935ed20cbe5)

PR #408, opened 2024-04-16
Run typecheck on CI

angelikatyborska:
This would have caught https://github.com/bitcrowd/tickety-tick/issues/406

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
     "lint": "eslint --ignore-path .gitignore --ext .js,.ts,.tsx .",
     "stylelint": "stylelint --ignore-path .gitignore 'src/**/*.scss'",
     "test": "jest",
+    "typecheck": "tsc --noEmit",
     "checks": "run-s stylelint lint test",
     "prepare-release": "./script/prepare-release",
     "release": "./script/release"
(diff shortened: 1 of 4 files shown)


klappradla:
Rebased this on `main` real quick 🏓

package.json
```diff
@@ -20,6 +20,7 @@
     "lint": "eslint --ignore-path .gitignore --ext .js,.ts,.tsx .",
     "stylelint": "stylelint --ignore-path .gitignore 'src/**/*.scss'",
     "test": "jest",
+    "typecheck": "tsc --noEmit",
```
klappradla, line 23: 💯 

Does this actually mean that we can drop some stuff from ESLint?

tsconfig.json
```diff
@@ -8,8 +8,9 @@
     "moduleResolution": "node",
     "noImplicitAny": true,
     "outDir": "./dist/",
+    "skipLibCheck": true,
```
klappradla, line 11: 👍 nice

klappradla (approved):
Great addition 🦄 

Wondering if we can slim down the ES-Lint stuff then 👀 ?

package.json
angelikatyborska: The typescript eslint plugin still has some useful checks in there (this is the list of default rules in the "recommended" config that we're using: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/configs/recommended.ts) so I would keep it

PR #410, opened 2024-04-17
Post release cleanups

klappradla:
Some cleanups after the rather rushed bugfix release:

- Add typecheck to pre-release checks
- Update tests for template-input component


angelikatyborska (approved):
Good idea 👍

PR #422, opened 2024-09-19
[311] Mozilla Add-ons manifest v3

### LINKED ISSUE #311: Mozilla Add-ons manifest v3
Check [Manifest v3 update](https://blog.mozilla.org/addons/2021/05/27/manifest-v3-update) whether we need to update the extension manifest.

klappradla:
This fixes #311 and migrates the extension to use [Manifest v3](https://developer.chrome.com/docs/extensions/develop/migrate/what-is-mv3). The most important change from v2 to v3 is that Firefox and Safari use background scripts or background pages while Chrome uses service workers for any background activity. They don't have a common ground though: Firefox & Safari extensions do not support background service workers and Chrome extensions do not support background pages or scripts.

So the respective `background` key in the `manifest.json` file accepts both, `scripts` and `service_worker` while the browsers only read what they can understand.

Tickety-Tick used a background script to load the ticket information from the current tab if there's any. Doing so it was using a global function assigned to the `window` object. Migrating to Manifest v3 we can't do this any more because Chrome's service workers don't have access to the DOM. The recommended mechanism of communication between the different parts of a browser extension (popup, content script, background script, etc.) are [messages](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/onMessage). So I updated our background script to use messages and work as both, a service worker and a background script.

Changes in detail:

* Update `web-ext` to v7.12.0
    
    This is the last version before the version change to v8.
    
    Update to v8 requires a bit more action eventually:
    https://github.com/mozilla/web-ext/releases/tag/8.0.0

* Update manifest to v3 for Firefox
    
    Firefox does not support background service workers. They still want
    scripts. So this setup still works fine in Firefox.

* Refactor background script
    
    Instead of relying on the background page instance (which is not
    recommended, see link to the docs), we communicate with the background
    script (or service worker script in Chrome) via message passing.
    
    This way, we can use the same script in both browsers.
    
    Mozilla docs on the topic: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Background_scripts#update_calls_for_background_script_functions), we

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,7 @@
     "stylelint-junit-formatter": "^0.2.2",
     "stylelint-prettier": "^4.0.2",
     "typescript": "^5.2.2",
-    "web-ext": "^7.6.0",
+    "web-ext": "7.12.0",
     "webpack": "5.94.0",
     "webpack-chain": "6.5.1",
     "webpack-cli": "^5.1.4",
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
     "bundle:chrome": "cross-env BUNDLE=true run-s build:chrome",
     "bundle:firefox": "cross-env BUNDLE=true run-s build:firefox",
     "lint": "eslint --ignore-path .gitignore --ext .js,.ts,.tsx .",
+    "format:js": "eslint --ignore-path .gitignore --fix --ext .js,.ts,.tsx .",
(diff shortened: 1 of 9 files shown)


andreasknoepfle (approved):
Neat stuff!

PR #423, opened 2024-09-24
Update release packages

klappradla:
This updates some of the packages involved for releasing the extension.

Changes:
* Upgrade web-ext dependency

* Update release script

  Use the new possibility to upload the human readable source code with the submission: https://extensionworkshop.com/documentation/develop/web-ext-command-reference/#whats-new

  For this to work, we need to download the Zip archive from GitHub first.

* Add lint commands for the extension

  This works as a sanity check for whether the manifest file we generate is correct. There's unfortunately currently a bug in Chrome which reports a false error….

* Update web extension polyfill

* Update chrome-launcher package

  This required to change the open-in-chrome script to an ES module.

* Update CLI for Chrome webstore uploads


hannahbit (approved):
Could not really test that it works, but looks good to me 🍰 

Thanks!

PR #424, opened 2024-09-24
Remove instructions for Opera

klappradla:
Opera users can just directly use the Chrome link - it has been long enough that Opera allows Chrome extensions.


hannahbit (approved):
LGTM 🍒

PR #425, opened 2024-09-24
[360] Check Safari build setup

### LINKED ISSUE #360: Check Safari build setup
> It didn't work, on my machine and Darren's too. Something with `xcodebuild` if I recall.

There seems to be an issue with the Safari build at the moment. Double check the setup, scripts and instructions. See [this comment](https://github.com/bitcrowd/tickety-tick/pull/352#issuecomment-1288543516) from @agatheblues.

klappradla:
Tiny doc-change for Safari instructions. The build works and the
installed extension is also persisted. The only thing that was necessary
for me was to install a new version of Java…

Fixes #360

### DIFF
--- a/SAFARI.md
+++ b/SAFARI.md
@@ -11,7 +11,7 @@ You’ll also need a signing certificate.
 The easiest way to create one might be this:
 
 1. Open Xcode
-2. Go to Preferences > Accounts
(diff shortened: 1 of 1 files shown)


hannahbit (approved):
Also tried building it and it worked for me. 👍

PR #430, opened 2024-10-12
feat: add support for plane

hng:
This PR adds an adapter to extract issue information from (self-hosted) [Plane](https://plane.so) instances. It extracts information from the url and the title tag and works from the issue detail page.

### DIFF
--- a/src/core/adapters/plane.test.ts
+++ b/src/core/adapters/plane.test.ts
@@ -11,7 +11,7 @@ const pages = {
     url: new URL(
       "https://app.plane.so/my_team/projects/e9460bd9-2967-4c5e-8d95-24b053e6965d/issues/9210a612-338b-4620-a834-07a89a26bf65/",
     ),
-    document: "<title>FOO-1 Do something interesting</title><meta property='og:title' content='Plane | Simple, extensible, open-source project management tool.'>",
+    document: "<title>FOO-1 Do something interesting</title><meta property='og:url' content='https://app.plane.so/'>",
   },
 };
 
--- a/src/core/adapters/plane.test.ts
+++ b/src/core/adapters/plane.test.ts
@@ -26,7 +26,7 @@ describe("plane adapter", () => {
     expect(result).toEqual([]);
   });
 
-  it("extracts the ticket from the issue page", async () => {
+  it("extracts the ticket info from the issue page", async () => {
     const result = await scan(
       pages.issuePage.url,
       doc(pages.issuePage.document),
--- a/src/core/adapters/plane.test.ts
+++ b/src/core/adapters/plane.test.ts
@@ -11,13 +11,16 @@ const pages = {
     url: new URL(
       "https://app.plane.so/my_team/projects/e9460bd9-2967-4c5e-8d95-24b053e6965d/issues/9210a612-338b-4620-a834-07a89a26bf65/",
     ),
-    document: "<title>FOO-1 Do something interesting</title><meta property='og:url' content='https://app.plane.so/'>",
+    document:
+      "<title>FOO-1 Do something interesting</title><meta property='og:url' content='https://app.plane.so/'>",
   },
 };
 
 describe("plane adapter", () => {
   function doc(head = "") {
-    const { window } = new JSDOM(`<html><head>${head}</head><body></body></html>`);
+    const { window } = new JSDOM(
+      `<html><head>${head}</head><body></body></html>`,
+    );
     return window.document;
   }
 
--- a/src/core/adapters/plane.test.ts
+++ b/src/core/adapters/plane.test.ts
@@ -1,0 +1,43 @@
+/**
+ * @jest-environment node
+ */
+
+import { JSDOM } from "jsdom";
+
+import scan from "./plane";
+
+const pages = {
+  issuePage: {
+    url: new URL(
+      "https://app.plane.so/my_team/projects/e9460bd9-2967-4c5e-8d95-24b053e6965d/issues/9210a612-338b-4620-a834-07a89a26bf65/",
+    ),
+    document: "<title>FOO-1 Do something interesting</title><meta property='og:title' content='Plane | Simple, extensible, open-source project management tool.'>",
+  },
+};
+
+describe("plane adapter", () => {
+  function doc(head = "") {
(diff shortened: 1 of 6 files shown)


src/core/adapters/plane.ts
```diff
@@ -0,0 +1,37 @@
...
+  let { issueType } = match("/:workspace/projects/:projectId/:issueType/:id/", url.pathname);
+  if (issueType == "issues") {
+    issueType = "feature";
+  }
+  const pageTitle = $text("title", document);
+  const extract = pageTitle && pageTitle.match(/([A-Z]+-\d+)\s(.+)/);
+  const id = extract && extract[1];
+  const title = extract && extract[2];
+
+  if (!id) return [];
```
klappradla, line 25: I think we can skip some checks and the extra return if we use the filtering in the end. So this should be enough:

```suggestion
  const pageTitle = $text("title", document);
  const extract = pageTitle?.match(/([A-Z]+-\d+)\s(.+)/);
  const id = extract && extract[1]
  const title = extract && extract [2];
```

But more importantly: Why is it necessary to extract the `ID` again, if it would already have been part of the URL?

src/core/adapters/plane.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  const ticket = {
+    id,
+    url: url.toString(),
+    title,
+    type: issueType,
+  } as TicketData;
```
klappradla, line 32: Let's align this with the other adapters and return a list that we filter later:

```suggestion
  const tickets = [{
    id,
    url: url.toString(),
    title,
    type: issueType,
  }];
```

src/core/adapters/plane.test.ts
```diff
@@ -0,0 +1,43 @@
...
+const pages = {
+  issuePage: {
+    url: new URL(
+      "https://app.plane.so/my_team/projects/e9460bd9-2967-4c5e-8d95-24b053e6965d/issues/9210a612-338b-4620-a834-07a89a26bf65/",
+    ),
+    document: "<title>FOO-1 Do something interesting</title><meta property='og:title' content='Plane | Simple, extensible, open-source project management tool.'>",
+  },
+};
+
+describe("plane adapter", () => {
```
klappradla, line 18: How about an additional test for a page that's part of the app **but not** the issue page? If there are any, I don't know the app.

klappradla (changes requested):
Hey @hng 👋 

thanks a ton for your contribution 🙏. Always interesting to see what other tools and trackers are out there...

The code looks generally fine. I left some tiny suggestions here and there. The for me personally most interesting part would be though: is there something distinctive about the issue pages which we could use as an early return from the `scan` function. I imagine that the app has more pages...

src/core/adapters/plane.ts
hng: The url includes a uuid and not the "issue id", I've changed the naming of the match variables.

src/core/adapters/plane.ts
```diff
@@ -0,0 +1,37 @@
...
+ * The adapter extracts ticket information from the markup.
+ */
+
+import { match } from "micro-match";
+
+import type { TicketData } from "../types";
+import { $text } from "./dom-helpers";
+
+async function scan(url: URL, document: Document): Promise<TicketData[]> {
+  const checkForPlane = document.querySelector("[property~='og:title'][content]");
```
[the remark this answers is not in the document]
hng: Sadly Plane does not includes a lot of hints for me to use. The application-name meta tag does not seem to exist on my self-hosted instance. I've changed it to look for the og:url tag.

Issues can also be opened in a preview slide over (similar to jira) but only the title on the detail issue page allows me to extract issue id and title as Plane frontends has zero hints in the HTML structure to extract issue id, title etc. (I'll may try to raise this as on issue on Plane)

src/core/adapters/plane.test.ts
hng: Yes there are also other pages, but I think checking the title should work well.

hng:
@klappradla Thanks for the suggestions! I've incorporated most of them. Sadly Plane has no hints in the HTML that we can use to extract issue id, title etc. so we are stuck with extracting them from the title for now.

src/core/adapters/plane.test.ts
```diff
@@ -0,0 +1,43 @@
...
+    const { window } = new JSDOM(`<html><head>${head}</head><body></body></html>`);
+    return window.document;
+  }
+
+  it("returns an empty array if it is on a different page", async () => {
+    const result = await scan(new URL("https://example.com"), doc());
+    expect(result).toEqual([]);
+  });
+
+  it("extracts the ticket from the issue page", async () => {
```
klappradla, line 29:
```suggestion
  it("extracts the ticket info from the issue page", async () => {
```

src/core/adapters/plane.ts
```diff
@@ -0,0 +1,37 @@
...
+import type { TicketData } from "../types";
+import { $find, $text } from "./dom-helpers";
+import { hasRequiredDetails } from "./utils";
+
+async function scan(url: URL, document: Document): Promise<TicketData[]> {
+  const ogUrl = <HTMLMetaElement> $find("[property~='og:url'][content]", document);
+  if (!ogUrl || ogUrl.content !== "https://app.plane.so/") return [];
+
+  let { issueType } = match("/:workspace/projects/:projectUuid/:issueType/:issueUuid/", url.pathname);
+  if (issueType == "issues") {
```
klappradla, line 18: Do other routes and other issue types even exist? If so, are they also plural in the path?

klappradla:
Thanks for the updates @hng 🙏 

Just some small questions around issue types left, then this should be good to go ⛵

src/core/adapters/plane.ts
hng: Yes but I can not test it as the community edition only has issues and having other issue types is a pro/enterprise feature that costs money...

I'd guess that these custom issue types should also be in plural, but as they be "free form" types

klappradla: Sounds good, thanks for explaining 🙏

hng:
@klappradla I did a rebase and fixed the format issues

klappradla:
Sweet! Thanks for your patience @hng 💚

Issue #448, opened 2025-06-10
GitLab Tickets broken


eheinle-mak:
We're making heavy use of this wonderful extension in our self hosted GitLab EE instance. With the recent UI changes, it does not recognize issues any more. The problems started back when switching back to the "old version" was still possible, however switching back to the old version did not fix the issue.

## Steps to reproduce

* Open GitLab Issue in its own browser tab
* Click the extension button
* Notice popup *No tickets found for this page*

## Expected behavior

* Popup shows options to copy branch names, commit messages or git commands

## Details

* Ubuntu 24.04, Google Chrome Version 137.0.7151.68 (Official Build) (64-bit)
* GitLab 18.0.1-ee

andreasknoepfle:
Hey @eheinle-mak,

I quickly added a patch for this. Do you feel capable of testing it locally before I release this with the steps outlined here on your local machine? https://github.com/bitcrowd/tickety-tick?tab=readme-ov-file#development . That would speed up the process :)

eheinle-mak:
Thanks for the quick fix and sorry for the late response from my side!

I can confirm it works now when I'm in the dedicated page for the issue. This is good enough for me and my workflow is restored.

Side notes though:

- Chrome warns about manifest v2 which will be/is deprecated
- It doesn't recognize a ticket when clicking an issue in GitLab's Kanban board, now that clicking an issue with left mouse button slides open an overlay with the issue contents instead of opening the issue page. I don't use that but I thought I'd mention it.

andreasknoepfle:
Released as v5.5.1

PR #449, opened 2025-06-10
[#448] GitLab Tickets broken

andreasknoepfle:


andreasknoepfle:
> Awesome!
> 
> Did you look into an issue's child items? I think those one are not being recognized either.

I think we didn't support that before yet, so I guess that is a new feature. 
We should also implement the modal view on the issues index eventually.

PR #453, opened 2025-10-03
Add support for Polarion

tiagotex:
I currently have to use [Polarion](https://polarion.plm.automation.siemens.com) 🥲 and miss being able to use Tickety-Tick to generate branches and commit messages.

I opted to get all information from the DOM instead of relying on the URL. I've tested this on my company's Polarion instance.

Thank you for the awesome plugin 💚

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,7 @@ Currently, we support:
 - [Tara](https://app.tara.ai/)
 - [Trello](https://trello.com/)
 - [YouTrack](https://www.jetbrains.com/youtrack/)
+- [Polarion](https://polarion.plm.automation.siemens.com)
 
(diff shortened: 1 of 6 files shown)


tiagotex:
Hey @andreasknoepfle nice to hear from you again too 😄 

We currently use the on-premises version of Polarion, but from what I gathered, they also offer a cloud version, so I opted to update the code to check for a Polarion class and return early if it's not present, this way it should still work even if using the cloud version.

I totally understand adding it as an unsupported adapter. I will keep an eye and update it if anything breaks. Do you want me to make any changes to the documentation to reflect this?

andreasknoepfle (approved):

PR #456, opened 2025-10-15
Fix chrome launcher

klappradla:
This updates the `open-in-chrome` script used for locally testing the extension in Chrome.

Chrome disabled the `--load-extension` flag some versions ago. The recommended way to load extensions now is via `--remote-debugging-pipe`. I updated the script accordingly, so that it works again. Before, it would just launch Chrome without loading any extension.

The code change is heavily inspired by the docs and tests from the `chrome-launcher` package and some AI help: https://github.com/GoogleChrome/chrome-launcher/blob/fe7654fbfd1ed651be9d576b6b46592b70e5b693/test/load-extension-test.ts

### DIFF
--- a/script/open-in-chrome.mjs
+++ b/script/open-in-chrome.mjs
@@ -2,22 +2,95 @@
 
 // usage: open-in-chrome [extension-dir] [starting-url]
 
-import * as path from 'path';
-import * as launcher from 'chrome-launcher';
+import * as path from "path";
+import * as launcher from "chrome-launcher";
 
 const dir = process.argv[2] || path.join(__dirname, "..", "dist", "chrome");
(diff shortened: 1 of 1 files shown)


script/open-in-chrome.mjs
```diff
@@ -2,22 +2,95 @@
...
+
+  chrome.process.on("exit", () => {
+    console.log("💨 Chrome closed.");
+    process.exit(0);
+  });
+}
+
+launchChrome().catch((err) => {
+  console.error("❌ Error:", err);
+  process.exit(1);
```
andreasknoepfle: Nice ⭐ 

We could maybe even extend this (in future PRs) to open tabs for all issue trackers that we have a publicly hosted example for, which would be absolutely great

PR #494, opened 2026-06-29
Patch critical and high severity vulnerabilities

klappradla:
I skipped one vulnerability, because it makes more sense to patch it in
`mdast-util-from-adf` instead.


planktonic:
I tried to review, but for some reason I get JS compile errors when I run `yarn build` on main. On this branch, it fails differently — it doesn’t even get to compiling the JS :/

joelpaulkoch:
`yarn build` fails for me too (looks like when it's building for chrome). I can however `yarn build:firefox` and `yarn build:chrome` individually 🤷 
I couldn't notice any issue with the firefox build.

klappradla:
Oh! Thanks @planktonic & @joelpaulkoch 🙏 

I just built and tested Firefox and Chrome. But it's probably the Safari build that's failing (it's based on Chrome). Will take a look.

klappradla:
It seems like the problem is only 1) updating XCode and then 2) getting the certificate right.

It does not have anything to do with the JS build. So I'd merge this PR and deal with Safari somewhere else.

